PR #1, opened 2023-01-16
Add form-preflight to plugin

pearofducks:
- Change to a monorepo structure to isolate deps and 'machinery'
- The `tw-reset.css` in the forms-preflight folder will eventually go away
- Once classes are wired up we should refactor index.html to take advantage of the plugin's CSS instead of manual `style`s
- Not super happy about `forms.js` (would prefer this being a CSS file), but this does make it so we don't need a spare build step to transform the CSS into a string - so meh?


nwikstrand (approved):
Given this runs as you intend it to, this looks good to me. At least for now :)

Just as ju mentioned already I would very much like it if CSS was in the form of "actual CSS" but that could be fixed later on. Not important at this point.

Good job Dave! 👏 🍐 🦆

PR #2, opened 2023-01-17
Wire up tw-reset preflight

pearofducks:
- Starts adding support for 'modes'
- The `hyper` mode is just a temporary name, but is the mode that assume zero NMP infrastructure (e.g. base.css, theme.css)

### DIFF
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -1,22 +1,25 @@
 // import { preflights } from './preflights.js'
 import { formPreflight } from '@nmp-ds/form-preflight'
+import { twReset } from './_preflights/tw-reset.js'
 import { rules } from '#rules'
 import { variants } from '#variants'
 import { theme } from './theme.js'
+
+const includePreflight = ['base', 'hyper']
 
 // export { theme, colors } from './theme.js'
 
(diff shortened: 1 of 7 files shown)


plugin/src/index.js
```diff
@@ -1,22 +1,25 @@
 // import { preflights } from './preflights.js'
```
nwikstrand, line 1: Might seem like a minor thing but I would move this to a separate file named `presetEngine.js` which would hold the current implementation, and have `index.js` file be responsible for exposing it to the outside world. Im not a fan of using `index`-files for actual implementations

```js
// plugin/src/presetEngine.js
function presetEngine(...) {
  ...
}

export default presetEngine;
```

```js
// plugin/src/index.js
import presetEngine from './presetEngine';

export { presetEngine }
```

nwikstrand (approved):
Looks good, would prefer if index files was used for exposing modules instead of holding implementations, but this is good to go as is.

Good job Dave 🥇 

68747470733a2f2f6d656469612e67697068792e636f6d2f6d656469612f6c324a4a724578396152736a4e727568692f67697068792e676966

PR #3, opened 2023-01-24
Add rules for inset and top, left, right bottom classes

nwikstrand:
This adds rules for various inset classes and top, left, right, bottom.

rules support the following, value suffix can of course differ, bounding will be added later.

```
inset-auto
inset-full
inset-1/2
inset-y-1
inset-x-1
top-1
left-1
right-1
bottom-1
```

All credit to @BalbinaK  for setting this up 🙌

### DIFF
--- a/plugin/src/_rules/position.js
+++ b/plugin/src/_rules/position.js
@@ -1,0 +1,24 @@
+import { handler as h, insetMap } from '#utils';
+
+function handleInsetValue(v, { theme }) {
+    return theme.spacing?.[v] ?? h.fraction.auto(v); // TODO: warn if no value
+}
+function handleInsetValues([, d, v], ctx) {
+    const r = handleInsetValue(v, ctx);
+    if (r != null && d in insetMap)
+        return insetMap[d].map(i => [i.slice(1), r]);
+}
+export const insets = [
+    [/^inset-(.+)$/, ([, v], ctx) => ({ inset: handleInsetValue(v, ctx) }),
+        {
+            autocomplete: [
+                '(inset-<directions>-$spacing',
+                'inset-(block|inline)-$spacing',
+                'inset-(bs|be|is|ie)-$spacing',
+                '(top|left|right|bottom)-$spacing',
+            ],
+        },
+    ],
+    [/^inset-([xy])-(.+)$/, handleInsetValues],
+    [/^(top|left|right|bottom)-(.+)$/, ([, d, v], ctx) => ({ [d]: handleInsetValue(v, ctx) })],
+];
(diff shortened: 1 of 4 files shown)


plugin/src/_rules/position.js
```diff
@@ -0,0 +1,24 @@
...
+    if (r != null && d in insetMap)
+        return insetMap[d].map(i => [i.slice(1), r]);
+}
+export const insets = [
+    [/^inset-(.+)$/, ([, v], ctx) => ({ inset: handleInsetValue(v, ctx) }),
+        {
+            autocomplete: [
+                '(inset-<directions>-$spacing',
+                'inset-(block|inline)-$spacing',
+                'inset-(bs|be|is|ie)-$spacing',
```
pearofducks, line 17: autocomplete needs cleaned up?

pearofducks (approved):
once autocomplete is cleaned up this lgtm

plugin/src/_rules/position.js
nwikstrand: Most def, will fix

nwikstrand: [be1322a](https://github.com/warp-ds/engine/pull/3/commits/be1322a797a93133bf77f44fb7343d2f83452d9a)

This ok or am I missing something, brain is tired, please halp

Skadefryd (approved):
Looks good to me :)

plugin/src/_rules/position.js
pearofducks: yep that's good!

nwikstrand:
Gaaah, forgot to squash the fixup 🤦

plugin/src/_rules/position.js
```diff
@@ -0,0 +1,23 @@
...
+    const r = handleInsetValue(v, ctx);
+    if (r != null && d in insetMap)
+        return insetMap[d].map(i => [i.slice(1), r]);
+}
+export const insets = [
+    [/^inset-(.+)$/, ([, v], ctx) => ({ inset: handleInsetValue(v, ctx) }),
+        {
+            autocomplete: [
+                '(inset-$spacing',
+                '(inset-<direction>-$spacing',
```
BalbinaK, line 16: @pearofducks Should we also allow autocomplete with "full", "auto", "1/2", "1/4"?

pearofducks: It seems like we might already be including that here? At least the original preset-mini autocomplete code doesn't have any special consideration for those?

I think it's a great thing to note down somewhere (a 'test autocomplete' task?) or keep in our mind for when we do check that though!

PR #4, opened 2023-01-25
feat: add rules for different grid and/or flex placements, alignments…

nwikstrand:
This adds rules for order, placement, alignments and justifies which makes it possible to generate styles for grid and flexbox positioning and order.

Paired with the almighty @BalbinaK 👑

fox

### DIFF
--- a/plugin/src/_rules/position.js
+++ b/plugin/src/_rules/position.js
@@ -1,4 +1,102 @@
 import { handler as h, insetMap } from '#utils';
+import { bounded } from "#bounding";
+
+
+const ORDER_BOUNDS = [1, 12];
+const numericHandler = { handler: (d) => h.number(d) };
+
+export const orders = [
+    [/^order-(\d+)$/,
+    bounded(
+      ([, d]) => ({ 'order': h.number(d)}),
+      ORDER_BOUNDS,
+      numericHandler
+    )],
+    ['order-first', { order: '-9999' }],
+    ['order-last', { order: '9999' }],
+    ['order-none', { order: '0' }],
+];
+
+export const justifies = [
+    // contents
+    ['justify-start', { 'justify-content': 'flex-start' }],
+    ['justify-end', { 'justify-content': 'flex-end' }],
+    ['justify-center', { 'justify-content': 'center' }],
+    ['justify-between', { 'justify-content': 'space-between' }],
+    ['justify-around', { 'justify-content': 'space-around' }],
+    ['justify-evenly', { 'justify-content': 'space-evenly' }],
+    // ...makeGlobalStaticRules('justify', 'justify-content'),
+
+    // items
+    ['justify-items-start', { 'justify-items': 'start' }],
+    ['justify-items-end', { 'justify-items': 'end' }],
+    ['justify-items-center', { 'justify-items': 'center' }],
+    ['justify-items-stretch', { 'justify-items': 'stretch' }],
+    // ...makeGlobalStaticRules('justify-items'),
+
+    // selfs
+    ['justify-self-auto', { 'justify-self': 'auto' }],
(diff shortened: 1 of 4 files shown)


plugin/src/_rules/position.js
```diff
@@ -1,4 +1,102 @@
...
+
+export const justifies = [
+    // contents
+    ['justify-start', { 'justify-content': 'flex-start' }],
+    ['justify-end', { 'justify-content': 'flex-end' }],
+    ['justify-center', { 'justify-content': 'center' }],
+    ['justify-between', { 'justify-content': 'space-between' }],
+    ['justify-around', { 'justify-content': 'space-around' }],
+    ['justify-evenly', { 'justify-content': 'space-evenly' }],
+    // ...makeGlobalStaticRules('justify', 'justify-content'),
```
pearofducks, line 28: These comments could either be nuked or with a TODO if there's still an action pending on them?

BalbinaK: We weren't sure how to approach this. Do you know who we should talk to if we need support for those global rules or if it makes sense to remove them?

pearofducks: With all these rules that aren't in Tailwind, I think it's fair we start with not including it and see if the community comes up with a valid and broad usecase.

Especially since these aren't in Fabric, and haven't been mentioned/requested!

Skadefryd: Does it matter that much if they are in Tailwind or not anymore as long as they seam useful?(/and not silly duplicates with just alias names) , we are officially leaving the "our thing is the same as Tailwind" -train now anyways with a full docsite and everything:)

Skadefryd (approved):
Looks good to me :) Nice work!    Insignificant, but why are some of those groups plural and some not.... like gap, padding, margin   vs    orders, justifies, alignments, insets   ?    order vs orders :D

plugin/src/_rules/position.js
pearofducks: Good point Snorre! Let's uncomment and keep them!

nwikstrand:
@Skadefryd @pearofducks will add tests for all these rules before this is merges so we'll request review again once that is done.

plugin/test/position.js
```diff
@@ -0,0 +1,29 @@
+import { getTestSuite } from './_helpers.js'
+import * as assert from 'uvu/assert'
+
+const test = getTestSuite('position')
+
+test('order allows values 1 to 12', async (t) => {
+  const range = [...Array(12).keys()].map(x => ++x)
```
pearofducks, line 7: Smallest possible nit - this 👇  feels a bit more clear than spreading `keys` into an array. Totally feel free to ignore, but I had to stop and think for a sec about what was going on there.

```
Array.from({ length: 12 }).map((_, i) => i + 1)
```

Note: `++i` would be fine here instead of `i + 1`, just used to people preferring the `i + 1` style 🤷

nwikstrand: Im not one of those thats against nitpicking, Im one of those that think that all of those small little nit picks actually adds to the greater good :) Will most dev look into your suggestion 🙏

nwikstrand:
Lots of repetition going on in tests which makes the test file very large in size and a bit heavy on the eye. Will look into that before merging this, should be fairly easy to do :)

pearofducks:
Sounds great! Would it make sense to consider snapshotting for some of these tests?

I normally am not a fan of snapshot tests when content might change often - but these really shouldn't _ever_ change once they're set up?

nwikstrand:
... hmm, didn't think of that. I don't use snapshot regularly. Lets look into it tomorrow, Im more than happy to rewrite this to something better 😵‍💫

PR #5, opened 2023-01-27
Scaffold vitest, migrate general.js as an example

pearofducks:


BalbinaK (approved):
Thanks for setting this migration up! Does it mean we can remove the commented bits in `general.js`?

pearofducks:
Yes we can!

I'll probably wait to merge this until #4 lands. I don't want to cause any merge conflicts (most of this can be copy/pasta over if needed)!

nwikstrand:
@pearofducks just merged #4 in order to not block this from going in. Lets gooooooo

PR #6, opened 2023-01-30
Align spacing to our use, tighten regexes, add tests

pearofducks:


nwikstrand (approved):
Good job Dave! 👏 Maybe it would be good to put some describe-blocks for padding and margin in there, just for structure, but this works as is :)

Just read up on the inline snapshots like 10 minutes ago :) pretty neat.

PR #8, opened 2023-01-30
Support global keywords in position rules

BalbinaK:
Add support for global keywords to rules in position.js, as suggested in a comment from https://github.com/warp-ds/engine/pull/4#discussion_r1086772020


nwikstrand (approved):
gmhjbyiaeqox6

PR #10, opened 2023-01-30
Feat: add rules for position, float, clear, z-index and box-sizing

BalbinaK:
This adds the rules to remaining position-related classes, i.e. position, float-, clear-, z- and box- classes, which makes it possible to generate respective styles.

### DIFF
--- a/plugin/src/_rules/position.js
+++ b/plugin/src/_rules/position.js
@@ -122,3 +122,17 @@ export const insets = [
     [/^inset-([xy])-(.+)$/, handleInsetValues],
     [/^(top|left|right|bottom)-(.+)$/, ([, d, v], ctx) => ({ [d]: handleInsetValue(v, ctx) })],
 ];
+
+export const floats = [
+  // floats
+  ['float-left', { float: 'left' }],
+  ['float-right', { float: 'right' }],
+  ['float-none', { float: 'none' }],
+  ...makeGlobalStaticRules('float'),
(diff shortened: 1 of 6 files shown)


plugin/src/_rules/position.js
```diff
@@ -119,3 +123,33 @@ export const insets = [
...
+  ...makeGlobalStaticRules('clear'),
+];
+
+function handleZIndexValue(v, { theme }) {
+  if (!theme.zIndex?.[v]) return warnOnce(`${v} is not allowed as z-index value`)
+  return theme.zIndex[v]
+}
+
+export const zIndexes = [
+  [/^z-(\d+)$/, ([, v], ctx) => ({ 'z-index': handleZIndexValue(v, ctx) }), { autocomplete: ['z-$zIndex', 'z-<num>'] }],
```
pearofducks, line 147: I don't think `z-<num>` should be in autocomplete? `z-$zIndex` is already covering what we allow for values.

pearofducks (approved):
Just the one change to autocomplete needed then this is looking great!

plugin/src/_rules/position.js
BalbinaK: Ah, good point! Removed `z-<num>` in [7aea540](https://github.com/warp-ds/engine/pull/10/commits/7aea5404f614ef99939988899568bd933119453b)

PR #11, opened 2023-01-31
Migrate all bounds to a single file

pearofducks:


plugin/src/index.js
```diff
@@ -24,3 +24,4 @@ export default presetWarp
 export { theme } from '#theme'
 export { rules } from '#rules'
 export { variants } from '#variants'
+export * as bounds from '#bounds'
```
BalbinaK, line 27: Why do we need to export bounds here, too?

BalbinaK (approved):

nwikstrand (approved):
giphy

plugin/src/index.js
pearofducks: So the docs (and maybe other places) can consume it

PR #18, opened 2023-02-02
Add space-x/y rules and space/divide variant

pearofducks:


plugin/src/_rules/space-margin.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const spaceMargins = [
+  [/^space-([xy])-(\d+)$/, handlerSpace, { autocomplete: ['space-(x|y)', 'space-(x|y)-reverse', 'space-(x|y)-$spacing'] }],
+  [/^space-([xy])$/, handlerSpace],
+  [/^space-([xy])-reverse$/, ([, d]) => ({ [`--w-space-${d}-reverse`]: 1 })],
+]
+
+function handlerSpace([, d, s = 1], { theme }) {
+  const v = theme.spacing?.[s]
+  if (v != null) {
```
nwikstrand, line 11: Not sure if this makes sense, but im thinking that `theme.spacing` is an object and targeting not existing items within that object will return `undefined`, not `null`, right?

```suggestion
  if (!Boolean(v)) {
```

nwikstrand (commented):
Great stuff 👏 good job Dave

plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@warp-ds/uno",
-  "version": "0.0.4",
+  "version": "0.0.5",
```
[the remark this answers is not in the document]
pearofducks: I think we need to worry about this when someone is actually consuming our software.

Pre v0 literally everything is considered a possible breaking change anyhow, so it's not a very big deal 😆

plugin/src/_rules/space-margin.js
pearofducks: Loose comparison to null catched both null and undefined

plugin/package.json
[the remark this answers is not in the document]
nwikstrand: Cool, Im good with that.. but if the version bump doesn't mean anything to us, just scrap it and go back to `0.0.0`.

plugin/src/_rules/space-margin.js
nwikstrand: Oki doki 😬 setting this to resolved

nwikstrand (approved):
Lägg ut! (I'll be impressed if you catch this reference)

dq1uph

plugin/package.json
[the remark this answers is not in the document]
pearofducks: The version bump means something to NPM 😆 

I'm publishing the package 'internally' for consumption in the CSS docs and elsewhere

plugin/src/_variants/spaceAndDivide.js
```diff
@@ -0,0 +1,10 @@
+export const variantSpaceAndDivide = (matcher) => {
+  if (/^space-?([xy])-?(-?.+)$/.test(matcher) || matcher.startsWith('divide-')) {
```
BalbinaK, line 2: Another question here is why can `([xy])` and the next group (`(-?.+)` be optional when the rules implemented in `_rules/space-margin` seem to always expect them? I mean that we do not support `space-2` or just `space` but always `space-x-2`, `space-y-144` or `space-y-reverse`

PR #19, opened 2023-02-02
feat: add rules for border and rounded

nwikstrand:
This adds rules for border and rounded classes, by the looks of it the added rules removes any related classes from the output of `checkFabricClasses.js`

### DIFF
--- a/plugin/src/_rules/border.js
+++ b/plugin/src/_rules/border.js
@@ -31,8 +31,7 @@ function handlerBorder(m, ctx) {
 }
 
 function handlerBorderSize([, a = "", b], { theme }) {
-  const v =
-    theme.lineWidth?.[b || "DEFAULT"] ?? h.bracket.cssvar.global.px(b || "1");
+  const v = theme.lineWidth?.[b ?? 2] ?? h.global(b);
   if (a in directionMap && v != null)
     return directionMap[a].map((i) => [`border${i}-width`, v]);
 }
@@ -51,7 +50,7 @@ export const rounded = [
 ]
 
 function handlerRounded([, a = '', s], { theme }) {
-  const v = theme.borderRadius?.[s || 'DEFAULT'] || h.bracket.cssvar.global.fraction.rem(s || '1');
+  const v = theme.borderRadius?.[s ?? 2] || h.global(s);
   if (a in cornerMap && v != null)
       return cornerMap[a].map(i => [`border${i}-radius`, v]);
 }
--- a/plugin/src/_rules/border.js
+++ b/plugin/src/_rules/border.js
@@ -1,0 +1,45 @@
+import { handler as h, directionMap, cornerMap } from "#utils";
+
+const borderStyles = [
+  "solid",
+  "dashed",
+  "dotted",
+  "double",
+  "hidden",
+  "none",
+  "groove",
+  "ridge",
+  "inset",
+  "outset",
+];
+
+export const borders = [
+  [/^border$/, handlerBorder],
+  [/^border()-(\d+)$/, handlerBorder, { autocomplete: "(border)-<directions>" }],
+  [/^border-([xy])-(\d+)$/, handlerBorder],
+  [/^border-([rltb])$/, handlerBorder],
+  [/^border-([rltb])-(\d+)$/, handlerBorder],
+  [/^border()-(.+)$/, handlerBorderStyle, { autocomplete: "(border)-<directions>" }],
+  [/^border-([xy])-(.+)$/, handlerBorderStyle],
+  [/^border-([rltb])-(.+)$/, handlerBorderStyle],
+];
+
+function handlerBorder(m, ctx) {
+  const borderSizes = handlerBorderSize(m, ctx);
+  const borderStyle = handlerBorderStyle(["", m[1], "solid"]);
+  if (borderSizes && borderStyle) {
+    return [...borderSizes, ...borderStyle];
+  }
+}
+
+function handlerBorderSize([, a = "", b], { theme }) {
+  const v =
+    theme.lineWidth?.[b || "DEFAULT"] ?? h.bracket.cssvar.global.px(b || "1");
+  if (a in directionMap && v != null)
+    return directionMap[a].map((i) => [`border${i}-width`, v]);
+}
+
+function handlerBorderStyle([, a = "", s]) {
+  if (borderStyles.includes(s) && a in directionMap)
+    return directionMap[a].map((i) => [`border${i}-style`, s]);
+}
--- a/plugin/src/_rules/border.js
+++ b/plugin/src/_rules/border.js
@@ -50,7 +50,7 @@ export const rounded = [
 ]
 
 function handlerRounded([, a = '', s], { theme }) {
-  const v = theme.borderRadius?.[s ?? 1] || h.global(s);
+  const v = theme.borderRadius?.[s ?? 4] || h.global(s);
   if (a in cornerMap && v != null)
       return cornerMap[a].map(i => [`border${i}-radius`, v]);
 }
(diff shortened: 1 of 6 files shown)


plugin/src/_rules/border.js
```diff
@@ -0,0 +1,59 @@
...
+  [/^border-([rltb])$/, handlerBorder],
+  [/^border-([rltb])-(\d+)$/, handlerBorder],
+  [/^border()-(.+)$/, handlerBorderStyle, { autocomplete: "(border)-style" }],
+  [/^border-([xy])-(.+)$/, handlerBorderStyle],
+  [/^border-([rltb])-(.+)$/, handlerBorderStyle],
+];
+
+function handlerBorder(m, ctx) {
+  const borderSizes = handlerBorderSize(m, ctx);
+  const borderStyle = handlerBorderStyle(["", m[1], "solid"]);
```
pearofducks, line 29: I think `handlerBorderStyle` might need tweaked or removed here.

Since we inherit [Tailwind's reset](https://github.com/warp-ds/drive/blob/master/plugin/src/_preflights/tw-reset.js#L14) we already have all borders set to be 'solid' - so we don't really want the 'bloat' on (f.ex) a `border-b-2` rule to include `border-style: solid`.

plugin/src/_rules/border.js
```diff
@@ -0,0 +1,59 @@
...
+  if (borderSizes && borderStyle) {
+    return [...borderSizes, ...borderStyle];
+  }
+}
+
+function handlerBorderSize([, a = "", b], { theme }) {
+  const v =
+    theme.lineWidth?.[b || "DEFAULT"] ?? h.bracket.cssvar.global.px(b || "1");
```
pearofducks, line 37: We probably want to tighten this down. In "theme" we should define `lineWidth` to just those few values we want to support (in pixels).

Then we can nuke the `h` call down to `h.global(b)` - that way we support those global values of `unset` and blah blah.

plugin/src/_rules/border.js
```diff
@@ -0,0 +1,59 @@
...
+export const rounded = [
+  [/^rounded()(?:-(.+))?$/, handlerRounded, { autocomplete: ['(rounded)', '(rounded)-<num>'] }],
+  [/^rounded-([rltb]+)(?:-(.+))?$/, handlerRounded],
+  [/^rounded([rltb]{2})(?:-(.+))?$/, handlerRounded],
+  [/^rounded-([bi][se])(?:-(.+))?$/, handlerRounded],
+  [/^rounded-([bi][se]-[bi][se])(?:-(.+))?$/, handlerRounded],
+]
+
+function handlerRounded([, a = '', s], { theme }) {
+  const v = theme.borderRadius?.[s || 'DEFAULT'] || h.bracket.cssvar.global.fraction.rem(s || '1');
```
pearofducks, line 56: Same here as `lineWidth`. Snag the border radii we supported in Fabric, throw those into theme, lock down the h-call to just `h.global`

plugin/test/border.js
```diff
@@ -0,0 +1,200 @@
...
+      .border-x-4{border-left-width:4px;border-right-width:4px;border-left-style:solid;border-right-style:solid;}
+      .border-x-8{border-left-width:8px;border-right-width:8px;border-left-style:solid;border-right-style:solid;}
+      .border-y-0{border-top-width:0px;border-bottom-width:0px;border-top-style:solid;border-bottom-style:solid;}
+      .border-y-2{border-top-width:2px;border-bottom-width:2px;border-top-style:solid;border-bottom-style:solid;}
+      .border-y-4{border-top-width:4px;border-bottom-width:4px;border-top-style:solid;border-bottom-style:solid;}
+      .border-y-8{border-top-width:8px;border-bottom-width:8px;border-top-style:solid;border-bottom-style:solid;}"
+    `);
+  });
+
+  test("supports setting border width", async (t) => {
```
pearofducks, line 28: Looks like missing a test for just `border` as a class (unless I missed it, it's super hard to search for) 🙈

nwikstrand: border-b, border-t coming up ⚡

plugin/test/border.js
```diff
@@ -0,0 +1,206 @@
...
+    const classes = ["border", "border-0", "border-2", "border-4", "border-8"];
+
+    const { css } = await t.uno.generate(classes);
+
+    expect(css).toMatchInlineSnapshot(`
+      "/* layer: default */
+      .border,
```
pearofducks, line 35: The shortform `.border` should give a 1px border

plugin/test/border.js
```diff
@@ -0,0 +1,206 @@
...
+
+describe("rounded", () => {
+  test("supports x|y with value", async (t) => {
+    const classes = ["rounded", "rounded-0", "rounded-2", "rounded-4", "rounded-8", "rounded-16", "rounded-full", "rounded-t", "rounded-t-0", "rounded-t-2", "rounded-t-4", "rounded-t-8", "rounded-t-16", "rounded-t-full", "rounded-r", "rounded-r-0", "rounded-r-2", "rounded-r-4", "rounded-r-8", "rounded-r-16", "rounded-r-full", "rounded-b", "rounded-b-0", "rounded-b-2", "rounded-b-4", "rounded-b-8", "rounded-b-16", "rounded-b-full", "rounded-l", "rounded-l-0", "rounded-l-2", "rounded-l-4", "rounded-l-8", "rounded-l-16", "rounded-l-full", "rounded-tl-0", "rounded-tl-2", "rounded-tl-4", "rounded-tl-8", "rounded-tl-16", "rounded-tl-full", "rounded-tr-0", "rounded-tr-2", "rounded-tr-4", "rounded-tr-8", "rounded-tr-16", "rounded-tr-full", "rounded-br-0", "rounded-br-2", "rounded-br-4", "rounded-br-8", "rounded-br-16", "rounded-br-full", "rounded-bl-0", "rounded-bl-2", "rounded-bl-4", "rounded-bl-8", "rounded-bl-16", "rounded-bl-full"];
+
+    const { css } = await t.uno.generate(classes);
+
+    expect(css).toMatchInlineSnapshot(`
+      "/* layer: default */
+      .rounded,
```
pearofducks, line 145: We should check with Adi what the 'default' `rounded` should give - we don't currently have a default in Fabric for this short-form.

pearofducks:
Looks great so far! A couple super minor things left.

pearofducks:
This continues to look good!

Was the final call here we support `.border` and friends for backward compat - but we don't do `.rounded` because reasons? (This is fine, just checking)

nwikstrand:
@pearofducks Im not sure, I settled for supporting both border and rounded and set them both to `1`.

pearofducks:
At least `.rounded` does not appear to be supported right now? Or if it is, it isn't tested.

I would use a default of 4 or 8 for `.rounded` if it is going in - `1` isn't used by anything and isn't one of our supported values that I know of.

nwikstrand:
> At least .rounded does not appear to be supported right now? Or if it is, it isn't tested.

In that case I must have broke something, will take a look at it.. again

nwikstrand:
> In that case I must have broke something, will take a look at it.. again

Yeah, turned out I didn't check the borderRadius object in `theme.js` after changing the value to `1`... key 1 isnt present in that object 🤦 

Imagine if we've used TS, that would have been catched. 

You may say Im a dreamer... but Im not the only one 🎶 👨‍🎤

pearofducks (approved):

pearofducks:
Sorry I approved this then just realized we have `h.global` here. We should either test these global values or drop that support. (I'm pretty sure `rounded-unset` wouldn't work right now looking at the code?)

pearofducks (approved):

PR #20, opened 2023-02-02
Add outline and shadow classes to DEPRECATIONS.md

BalbinaK:
This PR adds a markdown file we should populate with css classes that will probably no longer be supported in Warp Design System, or at least not in the same manner as in Fabric. We will then take a look at the full list and evaluate more in-depth how respective css rules should be handled.

Respective discussions on Slack:
- [about outline](https://sch-chat.slack.com/archives/C04LG5UTCTT/p1675342464480159)
- [about shadows](https://sch-chat.slack.com/archives/C04LG5UTCTT/p1675345388121489)

### DIFF
--- a/plugin/DEPRECATIONS.md
+++ b/plugin/DEPRECATIONS.md
@@ -1,0 +1,15 @@
+# These Fabric css rules will no longer be supported or need rethinking in Warp
+
+### Outline
+
+- outline-none
+- outline-white
+- outline-black
+
+### Shadow
+
+- shadow-2
+- shadow-3
+- shadow-4
+- shadow
+- shadow-none


nwikstrand (approved):
I know very little about the actual deprecations, but Im down with this being added. However, I could maybe see that it would be helpful if this was a section of `README.md` instead of a separate file so that it shows up in github whenever someone enteres the plugins-folder... possibly even that it should be moved up one level to be part of the non existing drive `README.md`

Anyways, Im down with this, feel free to disagree with the above and merge at will :)

pearofducks:
Great thought! I had intended the file as just for our reference when we go to make the docs  on this plugin - so it should die in a fire well before it hits 1.0 and public consumption.

PR #23, opened 2023-02-04
Rework project structure, get rid of monorepo

pearofducks:
Benefits:
- No more dependencies between modules, also no circular dependency
- Real Vite dev env available for testing classes
- Preflights all co-located


pearofducks:
closing in favor of #24 - an outstanding PR was open

PR #25, opened 2023-02-04
Moar rules

pearofducks:
Adds:
- transition
- transforms
- static
- behaviors

Adds to DEPRECATED:
- filters

We'll probably add custom rules later for box and drop shadows - but can easily do these outside of a larger 'filter' support like what TW or Uno has.


nwikstrand (commented):
Feel free to merge this if you feel safe about it, I've read through it and tried to check snapshots against rules but its hard to keep everything in mind due to the size of this PR.

src/_rules/grid.js
```diff
@@ -13,7 +13,8 @@ const autoDirection = (prop) => {
   // TODO: Emit warning if prop doesnt match
 };
 
-const numericHandler = { handler: (d) => h.number(d) }
+const numericHandler = { handler: (d) => h.number.auto(d) }
```
BalbinaK, line 16: `numericHandler` and `numericOrAutoHandler` seem to have the same statement. Was that intended?

src/_rules/static.js
```diff
@@ -0,0 +1,171 @@
...
+    { autocomplete: 'whitespace-(normal|nowrap|pre|pre-line|pre-wrap|break-spaces)' },
+  ],
+];
+
+// this is left flexible intentionally
+export const contents = [
+  [/^content-(.+)$/, ([, v]) => ({ content: h.bracket.cssvar(v) })],
+  ['content-empty', { content: '""' }],
+  ['content-none', { content: '""' }],
+];
```
BalbinaK, line 52: I see that in [unocss](https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/_rules/static.ts#L76-L82) we also have rules for `content-visibility`. I don't see it in Tailwind but it seems to be used in css in Finn code sometimes. Would that be a reason for us to add those rules?

pearofducks: Those are already supported with their TW/Fabric name [here](https://github.com/warp-ds/drive/blob/97973112eadfaf4feefa432467eda1e3a55d2eca/src/_rules/static.js#L3)

src/_rules/grid.js
pearofducks: Good catch!

src/_rules/static.js
BalbinaK: I asked because it looks like `content-visibility` is a separate CSS property [source](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility)

pearofducks: That's a very recent spec that isn't well supported.

PR #30, opened 2023-02-08
Replace local forks of variants with direct use of preset-mini

pearofducks:
Gotta love that `-344` ❤️


nwikstrand (approved):
Totally unrelated gif, but its as awesome as these changes.. lets go 🚀 

hound

test/__snapshots__/variants.js.snap
```diff
@@ -0,0 +1,41 @@
+// Vitest Snapshot v1
+
+exports[`variants > breakpoints 1`] = `
+"/* layer: default */
+@media (max-width: 767.9px){
+.lt-md\\\\:mt-32{margin-top:3.2rem;}
```
BalbinaK, line 6: I can't figure out why there are so many escape characters (`\\\\`) in vitest snapshots. It probably doesn't make any difference but just wondering what could be the reason. Is it something we did?

PR #32, opened 2023-02-08
Fix typo and update grid tests

AnnaRybkina:
Add tests for grid


test/grid.js
```diff
@@ -1,15 +1,116 @@
...
+    .grid-cols-4{grid-template-columns:repeat(4,minmax(0,1fr));}
+    .grid-cols-5{grid-template-columns:repeat(5,minmax(0,1fr));}"
+  `);
+});
+
+test("grid templates do not render if out of range", async (t) => {
+  const classes = ["grid-rows-8", "grid-rows-13"];
+  const { css } = await t.uno.generate(classes);
+
+  expect(css).toMatchInlineSnapshot('""');
```
AnnaRybkina, line 104: When running this test warnings get printed out in the console. I guess it is ok but just wanted to dubble check
image

AnnaRybkina: No warning was printed when running this test. They seem to be similar https://github.com/warp-ds/drive/blob/master/test/gap.js#L46

test/grid.js
```diff
@@ -1,15 +1,116 @@
...
+setup();
+
+test("grid span", async (t) => {
+  const classes = [
+    "col-auto",
+    "row-auto",
+    "col-span-full",
+    "row-span-full",
+    "row-span-5",
+    "col-span-3",
```
nwikstrand, line 13: Maybe we should test all values that are allowed within the bounds? 🤔 this would go for all classes with arbitrary values.

Alongside that we should probably also test that anything outside of the bounds doesn't generate any styling at all 💡

nwikstrand (commented):
Looks good to me, tho if am reading this right there seems to be some classes with arbitrary values missing in the tests. Might wanna look into that.

fox

nwikstrand:
Could also be an idea to generate snapshot files instead of inlineSnapshots just to have the test be less bloated.

src/_rules/grid.js
```diff
@@ -80,7 +80,7 @@ export const grid = [
   ['grid-flow-col', { 'grid-auto-flow': 'column' }],
   ['grid-flow-dense', { 'grid-auto-flow': 'dense' }],
   ['grid-flow-row-dense', { 'grid-auto-flow': 'row dense' }],
-  ['grid-flow-col-dense', { 'grid-auto-flow': 'col dense' }],
+  ['grid-flow-col-dense', { 'grid-auto-flow': 'column dense' }],
```
pearofducks, line 83: Good catch!

test/grid.js
AnnaRybkina: Yes and yes. I've added one test https://github.com/warp-ds/drive/pull/32/files#diff-1d3b2b3a27491b52437b1e12c9596a2a8f058641c4cf23c1d4455a8b14def09fR100 for the "out of bounds" test but we will add more of those. Makes total sense

AnnaRybkina: What do you think about this approach https://github.com/warp-ds/drive/pull/32/commits/c130ff22eb6635fea6831f90ce9d28172960609b @nwikstrand @pearofducks

src/_rules/grid.js
```diff
@@ -37,7 +41,7 @@ export const grid = [
       bounds.gridCol,
       numericHandler
     ),
-    { autocomplete: ['(row|col)-span-<num>'] }
+    { autocomplete: ['(row|col)-span-<num>'] },
```
pearofducks, line 44: Please turn off whatever lint/prettier is doing this (altering/fixing dangling-commas and syntax) - we definitely should configure a linter so we get consistent style, but for now this just makes PRs harder to read :)

test/grid.js
```diff
@@ -1,15 +1,126 @@
...
-  const { css } = await t.uno.generate([staticNumbers, arbitrary].flat(Infinity))
-  const staticResult = n => `grid-template-columns:repeat(${n},minmax(0,1fr));`
-  assert.include(css, staticResult(2))
-  assert.include(css, staticResult(5))
-  assert.include(css, 'grid-template-columns:200px minmax(900px, 1fr) 100px;')
-  assert.include(css, 'grid-template-columns:320px 1fr;')
-})
+import { setup } from './_helpers.js';
+import { expect, assert, test } from 'vitest';
+import { gridCol, gridRow } from '../src/bounds.js';
```
pearofducks, line 3:
```suggestion
import { gridCol, gridRow } from '#bounds';
```

test/grid.js
```diff
@@ -1,15 +1,126 @@
...
+    .grid-flow-row{grid-auto-flow:row;}
+    .grid-flow-col{grid-auto-flow:column;}
+    .grid-flow-dense{grid-auto-flow:dense;}
+    .grid-flow-row-dense{grid-auto-flow:row dense;}
+    .grid-flow-col-dense{grid-auto-flow:column dense;}"
+  `);
+});
+
+test('grid templates basic', async (t) => {
+  const numbers = Array.from({ length: 5 });
```
pearofducks, line 97: this isn't used is it?

AnnaRybkina: Yes, made some clean up now and added more tests.

test/grid.js
AnnaRybkina: Ah, ofc! tnx

src/_rules/grid.js
AnnaRybkina: alright. It is my VScode prettier. turning this off now

AnnaRybkina:
I **_think_** I've checked all of the cases now. Ready for re-review

PR #33, opened 2023-02-09
feat: add rules for table

nwikstrand:
This adds rules for various table related styles.

1. `border-collapse`, `border-separate`
2. `border-spacing-{spacing}`,`border-spacing-x-{spacing}`, `border-spacing-y-{spacing}` according to spacings defined in theme
3. `table-auto`, `table-fixed`


test/table.js
```diff
@@ -0,0 +1,113 @@
...
+})
+
+test('border-spacing', async (t) => {
+  const classes = spaceBase.map((spacingUnit) => {
+    return [`border-spacing-x-${spacingUnit}`, `border-spacing-y-${spacingUnit}`, `border-spacing-${spacingUnit}`];
+  }).flat()
+
+  const { css } = await t.uno.generate(classes)
+
+  expect(css).toMatchInlineSnapshot(`
```
pearofducks, line 38: Maybe a normal snapshot for something this long?

We could have a quick discussion if wanted on how long a 'snapshot' to prefer inline vs. not!

nwikstrand: Yeah, we just went with inline during the mob, but a regular one most def makes more sense :)

pearofducks (approved):

PR #35, opened 2023-02-14
Add theme opacity and tests

AnnaRybkina:
Background: Part of https://trello.com/c/JAeAbfHq/1-engine-add-rules-and-bounds

### DIFF
--- a/src/_rules/color.js
+++ b/src/_rules/color.js
@@ -1,15 +1,3 @@
-
-const handleOpacity = ([, d = ''], { theme }) => {
-  const v = theme.opacity[d];
-
-  if (v != null) {
-    return {
-        ['opacity']: v,
-    };
-}
-
-};
-
 export const opacity = [
-  [/^opacity-?(\d+)$/, handleOpacity, { autocomplete: 'opacity-${opacity}' }],
+  [/^opacity-?(\d+)$/, ([, d], { theme }) => ({ opacity: theme.opacity[d] }), { autocomplete: 'opacity-${opacity}' }],
 ];
--- a/src/_rules/color.js
+++ b/src/_rules/color.js
@@ -1,5 +1,15 @@
-import { handler as h } from '#utils';
+
+const handleOpacity = ([, d = ''], { theme }) => {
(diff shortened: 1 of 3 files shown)


AnnaRybkina:
[Engine: set theme values for opacity](https://trello.com/c/575Ys1FE/68-engine-set-theme-values-for-opacity)

src/_rules/color.js
```diff
@@ -1,5 +1,15 @@
-import { handler as h } from '#utils';
+
+const handleOpacity = ([, d = ''], { theme }) => {
```
pearofducks, line 2: Have you checked if this can be handled down in the tuple as this shortened version?

```
([, d]) => ({ opacity: theme.opacity[d] })
```

pearofducks: (Had a free moment so I checked this and all tests pass!)

src/_rules/color.js
```diff
@@ -1,5 +1,15 @@
...
+    return {
+        ['opacity']: v,
+    };
+}
+
+};
 
 export const opacity = [
-  [/^opacity-?(\d+)$/, ([, d]) => ({ opacity: h.bracket.percent(d) })],
+  [/^opacity-?(\d+)$/, handleOpacity, { autocomplete: 'opacity-${opacity}' }],
```
pearofducks, line 14:
```suggestion
  [/^opacity-?(\d+)$/, ([, d], { theme }) => ({ opacity: theme.opacity[d] }), { autocomplete: 'opacity-${opacity}' }],

```

src/_rules/color.js
pearofducks, line 2: Delete this in favor of the change proposed below?

src/_rules/color.js
AnnaRybkina: Great suggestion. I fix 👍

src/_rules/color.js
AnnaRybkina: Yes, fixing

pearofducks (approved):

PR #36, opened 2023-02-14
Add background rules

pearofducks:


test/__snapshots__/background.js.snap
```diff
@@ -0,0 +1,41 @@
...
+.bg-origin-content{background-origin:content-box;}"
+`;
+
+exports[`bg positions 1`] = `
+"/* layer: default */
+.bg-bottom{background-position:bottom;}
+.bg-center{background-position:center;}
+.bg-left{background-position:left;}
+.bg-right{background-position:right;}
+.bg-top{background-position:top;}"
```
AnnaRybkina, line 23: It seems that we do not support "nested" positions like `bg-left-bottom`, `bg-left-top`, `bg-right-bottom`, `bg-right-top`. Is it intentional?

pearofducks: @AnnaRybkina - good catch, fixed!

AnnaRybkina (approved):
🚀

PR #39, opened 2023-02-16
Backport aspect ratio, set up the first internal classes

pearofducks:
- Add aspect ratio in a more ergonomic and TW-ish way: e.g. `aspect-16/9` instead of `aspect-w-16 aspect-h-9`
- Drop support for `aspect-none`, only one team was using it - the same effect should be doable using our improved support for breakpoints - e.g. `lt-sm:aspect-16/9` instead of `aspect-16/9 sm:aspect-none`
- Set up first internal classes, not sure if it might be better to include what they target, but we can fix that later - e.g. we might want `i-text-c-$foo` so we know it targets color instead of just `i-text-$foo`. Because we're only reading straight from a CSSvar with all these rules we can't really make 'smart' rules that change behavior based on some pattern.

### DIFF
--- a/src/_utils/handlers/handlers.js
+++ b/src/_utils/handlers/handlers.js
@@ -146,7 +146,7 @@ export function bracketOfLength(str) {
 export function bracketOfPosition(str) {
     return bracketWithType(str, 'position');
 }
-export function warptoken(str) {
+export function warpToken(str) {
     if (str.match(/^\$\S/))
         return `var(--w-${escapeSelector(str.slice(1))})`;
 }
--- a/src/_utils/handlers/handlers.js
+++ b/src/_utils/handlers/handlers.js
@@ -70,6 +70,14 @@ export function percent(str) {
     if (!Number.isNaN(num))
         return `${round(num / 100)}`;
 }
+export function inverseFraction(str) {
+    if (str === 'full')
+        return '100%';
+    const [left, right] = str.split('/');
+    const num = parseFloat(right) / parseFloat(left);
+    if (!Number.isNaN(num))
+        return `${round(num * 100)}%`;
+}
 export function fraction(str) {
     if (str === 'full')
         return '100%';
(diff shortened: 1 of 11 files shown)


src/_utils/handlers/handlers.js
```diff
@@ -138,6 +146,10 @@ export function bracketOfLength(str) {
 export function bracketOfPosition(str) {
     return bracketWithType(str, 'position');
 }
+export function warpToken(str) {
+    if (str.match(/^\$\S/))
+        return `var(--w-${escapeSelector(str.slice(1))})`;
+}
```
AnnaRybkina, line 152: As this function is very much alike the `cssvar()` besides the `--w` part those two can be refactored and maybe the matching part can be moved to a separate function. But now when I'm writing that I'm not sure if it would improve readability for example. Feel free to disregard my comment but I'll leave it here fwiw

test/aspect-ratio.js
```diff
@@ -0,0 +1,10 @@
+import { setup } from './_helpers.js'
+import { expect, test } from 'vitest'
+
+setup()
+
+test('it generates backported aspect-ratios', async ({ uno }) => {
+  const classes = ['aspect-16/9', 'aspect-1/1', 'aspect-4/3', 'aspect-2/1']
+  const { css } = await uno.generate(classes)
+  expect(css).toMatchSnapshot()
```
AnnaRybkina, line 9: Would it make sense to add a test for the values that we do not support to test that we do not generate any styling at all then?

test/internal.js
```diff
@@ -0,0 +1,11 @@
+import { setup } from './_helpers.js'
+import { expect, test } from 'vitest'
+
+setup()
+
+test('it generates css based on warp tokens', async ({ uno }) => {
+  const classes = ['i-bg-$foo-bar', 'i-text-$biz-baz', 'i-border-$wombat-llama']
+  const anticlasses = ['i-bg-foos-bars', 'i-text-bizs-bazs', 'i-border-wombats-llamas', 'i-magnuseses']
```
AnnaRybkina, line 8: Love it `i-magnuseses` 😁

AnnaRybkina (approved):
Great stuff 🙌 I left some comments but I approve as the changes are awesome and you can ignore the comments as they were mostly thoughts out loud

PR #41, opened 2023-02-16
Linting

pearofducks:


nwikstrand (approved):
Lint setup looks fine (have not scrolled through the rest). Would be cool if we run the lint pre-push or commit 💡

package.json
```diff
@@ -23,7 +23,8 @@
     "dev:env": "vite dev-env --config ./vite.config.js",
     "test": "vitest",
     "build": "rollup -c",
-    "version": "pnpm test && pnpm build && pnpm publish"
+    "version": "pnpm test && pnpm build && pnpm publish",
+    "lint": "eslint . --fix"
```
AnnaRybkina, line 27: What do you think about this suggestion? Just to be clear that you are not only running lint but also fixing the issues? I do not have a strong opinion here though.

```suggestion
    "lint:fix": "eslint . --fix"
```

pearofducks:
@nwikstrand @AnnaRybkina - I think both of your suggestions make sense, possibly as a follow-up PR to set this up in CI or do whatever the team decides is best. I don't have strong opinions on when/how linting happens, so I leave it up to y'all.

PR #42, opened 2023-02-16
Add list-checked rule and corresponding test

magnuh:
Added list-checked class from here:
https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/list.css

Question: Should we add fallbacks to all vars.

### DIFF
--- a/src/_rules/list.js
+++ b/src/_rules/list.js
@@ -1,0 +1,28 @@
+import { entriesToCss } from '@unocss/core';
+
+const listItemStyles = entriesToCss(Object.entries({
+  position: 'relative',
+  'padding-left': 'var(--w-space-3)',
+}));
+const checkmarkStyles = entriesToCss(Object.entries({
+  content: '""',
+  display: 'block',
+  position: 'absolute',
+  width: '16px',
+  height: '28px',
+  left: 0,
+  color: 'var(--w-list-color-icon-checked)',
+  'background-size': 'contain',
+  'background-position': '50%',
+  'background-repeat': 'no-repeat',
+  'background-image': '""',
+}));
+
+export const listChecked = [
+  [/^list-checked$/, ([selector]) => {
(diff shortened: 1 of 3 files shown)


src/_rules/list.js
```diff
@@ -0,0 +1,28 @@
...
+  'padding-left': 'var(--w-space-3)',
+}));
+const checkmarkStyles = entriesToCss(Object.entries({
+  content: '""',
+  display: 'block',
+  position: 'absolute',
+  width: '16px',
+  height: '28px',
+  left: 0,
+  color: 'var(--w-list-color-icon-checked)',
```
magnuh, line 14:
```suggestion
  color: 'var(--w-list-color-icon-checked, hotpink)',
```
Should we be adding fallback to all variables?

pearofducks: No fallback needed in any of these rules IMO

src/_rules/list.js
```diff
@@ -0,0 +1,28 @@
+import { entriesToCss } from '@unocss/core';
+
+const listItemStyles = entriesToCss(Object.entries({
+  position: 'relative',
+  'padding-left': 'var(--w-space-3)',
```
pearofducks, line 5: I don't think the `space` values are ready for use?

magnuh: Perhaps not... 😄  How should we handle it for now?

pearofducks: I would just set it to static pixels for now like the heights and widths.

magnuh: Should we use "TODOs" in these places? Or would that make it too todo cluttered at this time? 😅

magnuh: Updated with todos.

pearofducks (approved):

AnnaRybkina:
Great stuff! Merging 🚢

PR #43, opened 2023-02-17
Tests for display and line-clamp

AnnaRybkina:
- [x] Moved out some big inline snapshots in favor of snapshot files(position.js)


AnnaRybkina:
[Add missing tests for display and line-clamp](https://trello.com/c/5XF1uopV/77-add-missing-tests-for-display-and-line-clamp)

test/line-clamp.js
```diff
@@ -0,0 +1,33 @@
...
+
+
+setup();
+
+describe('line-clamp', () => {
+  test('should render styles for static classes', async ({ uno }) => {
+    const staticClasses = getNumericArrayInRange(1, 9).map(n => `line-clamp-${n}`);
+
+    const { css } = await uno.generate(staticClasses);
+    expect(css).toMatchInlineSnapshot(`
```
pearofducks, line 13: Why inline snapshot here and removing all inline snapshot in `position.js`? Many of those inline snapshots there are shorter than this one. 😕

test/display.js
```diff
@@ -0,0 +1,34 @@
+import { setup } from "./_helpers.js";
+import { describe, expect, test } from "vitest";
+import * as displayRules from '../src/_rules/display.js';
```
nwikstrand, line 3: Isnt this exposed from _rules index? would be cleaner to import from there instead, if possible 💡 

```suggestion
import * as displayRules from '../src/_rules/display.js';
```

test/line-clamp.js
pearofducks: OK! 

To be clear if I post a question I mean it as a genuine question, there might be a good for something that I don't understand - so please don't feel like some immediate action is needed!

test/display.js
pearofducks: Yep, this should be able to be `#rules` for the import - [defined here](https://github.com/warp-ds/drive/blob/master/package.json#L14)

test/line-clamp.js
AnnaRybkina: Right, got it.

test/display.js
AnnaRybkina: Maybe we can fix those imports in a better matter in a separate PR? Another example is here https://github.com/warp-ds/drive/blob/master/test/static.js#L3

pearofducks (approved):

PR #44, opened 2023-02-17
Wire up externalization of 'core' classes

pearofducks:
This creates support for an array of classes (like our component classes) to be passed to the plugin and then blacklisted from the resulting CSS.

### DIFF
--- a/src/postprocessor.js
+++ b/src/postprocessor.js
@@ -2,13 +2,14 @@ import { escapeSelector } from '@unocss/core'
 
 const prefix = 'w-'
 const classify = (str) => `.${escapeSelector(str)}`
-const coreClasses = [].map(classify) // temporary fixture
 
-export const postprocess = (externalizeCore) => (obj) => {
-  if (externalizeCore && coreClasses.includes(obj.selector)) return obj.entries = null
-  obj.entries.forEach(e => {
-    e[0] = e[0].replace(/^--un-/, `--${prefix}`);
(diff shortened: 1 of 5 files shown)


src/postprocessor.js
```diff
@@ -0,0 +1,15 @@
...
+
+const prefix = 'w-'
+const classify = (str) => `.${escapeSelector(str)}`
+
+export const postprocess = (shouldExternalize, _externalClasses = []) => {
+  const externalClasses = _externalClasses.map(classify)
+  return (obj) => {
+    if (shouldExternalize && externalClasses.includes(obj.selector)) return obj.entries = null
+    obj.entries.forEach(e => {
+      e[0] = e[0].replace(/^--un-/, `--${prefix}`);
```
BalbinaK, line 11: It looks like `--un-` comes from `colorResolver` util. Would it make sense to change that prefix to `w-` in that util?

BalbinaK (approved):
Just a small comment.

src/postprocessor.js
pearofducks: This function will catch all occurances of `--un`, so we don't need to worry about if it exists in some files (f.ex we have externalized most variants, so it could be happening there too!).

BalbinaK: Would it still make sense to update the prefix in our util? Just checking :)

pearofducks: Doesn't matter either way I don't think?

PR #46, opened 2023-02-20
Restrict arbitrary values for display

AnnaRybkina:
@BalbinaK raised an interesting question about the reason for supporting whatever value to be provided as an arbitrary value for the display property. This PR removes fallback to any value provided for display and improves tests and rule a bit.
Let's discuss and see if this change make sense 🙌

### DIFF
--- a/src/_rules/display.js
+++ b/src/_rules/display.js
@@ -24,5 +24,5 @@ export const display = [
   ['table-row', { display: 'table-row' }],
   ['table-row-group', { display: 'table-row-group' }],
   // unset, revert, inherit
-  [/^display-(.+)$/, ([, c]) => ({ display: h.cssvar.global(c) }), { autocomplete: `display-(${globalKeywords.join('|')}` }],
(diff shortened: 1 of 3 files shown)


BalbinaK (approved):

src/_rules/display.js
```diff
@@ -23,5 +24,5 @@ export const display = [
   ['table-row', { display: 'table-row' }],
   ['table-row-group', { display: 'table-row-group' }],
   // unset, revert, inherit
-  [/^display-(.+)$/, ([, c]) => ({ display: h.bracket.cssvar.global(c) || c })],
+  [/^display-(.+)$/, ([, c]) => ({ display: h.cssvar.global(c) }), { autocomplete: `display-(${globalKeywords.join('|')}` }],
```
AnnaRybkina, line 27:
```suggestion
  [/^display-(.+)$/, ([, c]) => ({ display: h.cssvar.global(c) }), { autocomplete: `display-${globalKeywords.join('|')}` }],
```

PR #48, opened 2023-02-21
Add support for providing options from terminal

AnnaRybkina:
Before we could run `node dev.js 'gap-2'` to get the output like(default to rem) 👇 
```
/* layer: default */
.gap-2{gap:0.2rem;}
```

We still can do the above but now we can also provide an option with `--` prefix to use it in the theme. More info in the README
Like that(use pixels instead of rem) -> `node dev.js --usePixels --cliClasses '-m-2! gap-2'`
```
/* layer: default */
.gap-2{gap:2px;}
```

More usages: `node dev.js --usePixels --development --cliClasses '-m-2! gap-2'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,11 @@
 1. Check out the `parity` project from warp-ds, get dependencies for the project
 2. Link to the `parity` project from the `plugin` folder (this folder): `pnpm link ../parity`
 3. Run `node checkFabricClasses.js`
+
+## generating warp classes using command line
+
+Run `node dev.js` or `pnpm dev`
+  Usage:
+    node dev.js [-c <string> | --cliClasses=<string>] [--usePixels] [--development] [--externalClasses] [--externalizeClasses] [--usePreflight]
+    Example: `node dev.js --usePixels --development --cliClasses=m-2`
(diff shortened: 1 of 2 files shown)


pearofducks:
https://nodejs.org/api/util.html#utilparseargsconfig

Another option is to require Node 18 (which is the `current` version of Node) and use the built in util.

If we're making improvements to the dev script, it would be nice to see it able to handle negative values - IIRC `-mb-32` doesn't work today. Ofc this can be handled in a separate PR if wanted.

BalbinaK:
> IIRC `-mb-32` doesn't work today.

@pearofducks As far as I have tested, negative values work both when running `pnpm dev -m-12` and `node dev.js -m-12`.

AnnaRybkina:
> https://nodejs.org/api/util.html#utilparseargsconfig
> 
> Another option is to require Node 18 (which is the `current` version of Node) and use the built in util.
> 
> If we're making improvements to the dev script, it would be nice to see it able to handle negative values - IIRC `-mb-32` doesn't work today. Ofc this can be handled in a separate PR if wanted.

What do you think of [this](https://github.com/warp-ds/drive/pull/48/commits/a815e889043b2a0ff533fb0a0fae0e9b5989736c)

PR #50, opened 2023-02-21
Add a linting check to PRs

pearofducks:
Prevent the need for any further lint-fix PRs by blocking anything from going in that isn't nice and tidy!

### DIFF
--- a/.github/workflows/lint.pr.yml
+++ b/.github/workflows/lint.pr.yml
@@ -1,0 +1,18 @@
+name: Lint
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: current
+    - uses: pnpm/action-setup@v2
+      with:
+        version: 7
+        run_install: true
+    - name: Lint
+      run: pnpm lint:check
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "test": "vitest",
     "build": "rollup -c",
     "version": "pnpm test && pnpm build && pnpm publish",
-    "lint": "eslint . --fix"
+    "lint": "eslint . --fix",
+    "lint:check": "eslint ."
   },
   "keywords": [],
   "author": "@djh",


.github/workflows/lint.pr.yml
```diff
@@ -0,0 +1,18 @@
...
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: current
```
BalbinaK, line 12: Just curious, what's the reason for choosing `current` node version here? In the [actions/setup-node](https://github.com/actions/setup-node#usage) readme they say `"it is recommended to always specify Node.js version and don't rely on the system one."`

.github/workflows/lint.pr.yml
```diff
@@ -0,0 +1,18 @@
+name: Lint
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
```
BalbinaK, line 10: Another minor comment. Would it make sense to set names for those actions?
E.g.    
```
- name: Checkout
  uses: actions/checkout@v3
...
- name: Install Node.js
  uses: actions/setup-node@v3
...
- name: Install pnpm
  uses: pnpm/action-setup@v2

BalbinaK (approved):
Thanks for adding that! Just small comments, as usual :)

And a question: do we need to add something to the yml file to make this job ✅ required before merging a PR?

.github/workflows/lint.pr.yml
pearofducks: Nah, it gives them clear enough names, and those things will very rarely fail.

pearofducks:
I think in the GitHub controls we can specify that "all checks must pass" before a PR merges? That'd be the only thing we want to set IMO - but I'd be surprised if someone merges a failing PR 🤷

.github/workflows/lint.pr.yml
BalbinaK: I see. Just to understand what could happen in the future with this. Can using “current” mean that if there’s a breaking change in node that affects our code in warp drive, the job will fail forcing us to either update the scripts or change the node version to what we’ve supported so far?

.github/workflows/lint.pr.yml
BalbinaK: Based on what comes after “uses:” I actually had to do some research to understand what’s happening, because I’m not yet that familiar with GitHub actions. I thought others could have a similar problem in the future but maybe that’s a good learning point.

.github/workflows/lint.pr.yml
pearofducks: Yes, but this is a good thing. We *want* our tests to fail when a new version of Node would cause something to break - that way we find out about it before our users do!

BalbinaK: Correct me if I'm wrong but I thought that specifying node version within a module is only relevant to that module and not other apps that use that module. Then if we expect the module to work on Node 18 per today and a github action fails in a year from now because it's using node 20, then my understanding was that we just affect ourselves.

pearofducks: That sounds right?

BalbinaK: 😄 Then I didn't get what you meant by that comment:
> We want our tests to fail when a new version of Node would cause something to break - that way we find out about it before our users do

PR #60, opened 2023-03-01
Get rid of bounding

pearofducks:


Skadefryd (approved):

test/flex.js
```diff
@@ -62,7 +59,7 @@ test('flex invalid', async (t) => {
 });
 
 test('flex shrink/grow', async (t) => {
-  const boundsRangeArray  = getNumericArrayInRange(flexGrowShrink[0], flexGrowShrink[1]);
+  const boundsRangeArray  = getNumericArrayInRange(0, 5);
```
BalbinaK, line 62: Could you rename the constant - it was a bit confusing with the PR title :)

src/_utils/utilities.js
```diff
@@ -10,11 +10,9 @@ import { getComponents } from './getComponents.js';
  * @param {string} propertyPrefix - Property for the css value to be created. Postfix will be appended according to direction matched.
  * @see {@link directionMap}
  */
-// TODO: hook this into the 'bounding' utils
 export function directionSize(propertyPrefix) {
   return ([_, direction, size], { theme }) => {
-    const v = theme.spacing?.[size] ?? h.global.auto(size);
-    if (!v) return warnOnce(`${propertyPrefix} not available in size ${size}`);
+    const v = theme.spacing?.[size] ?? h.bracket.global.auto.fraction(size);
```
BalbinaK, line 15: Does it mean that a classname `m-13` will be accepted? Or would it need to be `m-[13]`?

BalbinaK (commented):
@pearofducks @Skadefryd It would be helpful with some PR description, even though it might be obvious to some based on the code changes. That or a rough summary of the changes helps me in the review process.

What about border-radius, opacity, border and anything else that relies on bounds specified in `theme.js`? Should those be updated to accept arbitrary values, too?

test/flex.js
AnnaRybkina: Addressed here https://github.com/warp-ds/drive/pull/62/files#diff-4523a55c502753c1e3fbb6cd0db39d893b95f593d8166e4822dd76ffc005247bR63

PR #67, opened 2023-03-10
Update license in package.json

AnnaRybkina:
Update to the correct license type https://github.com/warp-ds/drive/blob/master/LICENSE#L1

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   },
   "keywords": [],
   "author": "@djh",
-  "license": "MIT",
+  "license": "Apache 2.0",
   "dependencies": {
     "@unocss/core": "^0.50.4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
   },
   "keywords": [],
   "author": "@djh",
-  "license": "MIT",
+  "license": "Apache 2.0",
```
pearofducks, line 31:
```suggestion
  "license": "Apache-2.0",
```

This must be a valid value from the [SPDX License List](https://spdx.org/licenses/)

AnnaRybkina: 🤦‍♀️

magnuh (approved):

PR #68, opened 2023-03-10
Automate package publishing using semantic-release 

BalbinaK:


BalbinaK:
[Automate releases of warp packages to NPM and EIK](https://trello.com/c/v1Ptqg8G/128-automate-releases-of-warp-packages-to-npm-and-eik)

package.json
```diff
@@ -17,12 +18,15 @@
     "#preflights": "./src/_preflights/index.js",
     "#postprocess": "./src/postprocessor.js"
   },
+  "files": [
+    "dist"
```
pearofducks, line 22: I'm pretty sure this would cause some of the specified `exports` to fail.

BalbinaK: Good point. Do we want everything from the exports entry to be published to NPM? Right now all repo contents are available there, which doesn't sound necessary. What do you think?

pearofducks: I think we can use an `.npmignore` file to remove our test files - otherwise I think everything else is used.

BalbinaK: I see you tried adding a rule `!dist/` to `. npmignore`, which sounds to me as if you wanted to publish just the dist folder after all. That rule didn't work so we published everything. Do we indeed want to just exclude the tests folder? Just trying to connect the dots :)

BalbinaK: Instead of adding more to `.npmignore` I was thinking to set `src`, `dist` and `dev.js` as files published to NPM. Wdyt? [f415da5](https://github.com/warp-ds/drive/pull/68/commits/f415da56de51dfb431c3a46428ef6852841dabcc)

pearofducks: That works!

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@warp-ds/uno",
```
magnuh, line 2: Been wondering about this... is there a reason why the package is named `uno` and not `drive`? 🤔😅

pearofducks: Users have no relation to the word `drive`. They're using this package in an UnoCSS context. We don't need to be filling users' heads with "OK I want the Warp Uno preset, was that `@warp-ds/drive` or `@warp-ds/speed` or the other gimmicky package name they went with?" 😆 

`preset-uno` might be an even better name - but it's possible at some point we want to export more than just a preset from this package.

magnuh: ...then my next question is: Why is it called `drive` and not `uno` in github? 🤷‍♂️😅😜  I was a bit confused about this since all other repos is mapped 1:1 between github and npm? (Or at least I think they are? 🤔😄)

pearofducks: Basically "for the lulz". I agree with your logic, we could have some kind of alias repo to be sure people find the right thing. I'm up for whatever the team feels is best here to aid users understanding our packages and being able to discover things.

PR #73, opened 2023-03-17
fix: bad comment syntax in preflight

pearofducks:
- Bumps deps, specifically Vite 4.2 which adds support for subpath imports
- Fixes a bad comment syntax in the preflight


AnnaRybkina (approved):
Good catch

PR #76, opened 2023-03-27
fix: add bg color options 

AnnaRybkina:
Added more bg color options as they are needed for e.g. button component

### DIFF
--- a/src/_rules/color.js
+++ b/src/_rules/color.js
@@ -12,4 +12,7 @@ export const textColors = [
 export const bgColors = [
   ['bg', { 'background-color': 'var(--w-background-color)' }],
   ['bg-subtle', { 'background-color': 'var(--w-background-color-subtle)' }],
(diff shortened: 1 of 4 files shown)


src/_rules/color.js
```diff
@@ -12,4 +12,7 @@ export const textColors = [
 export const bgColors = [
   ['bg', { 'background-color': 'var(--w-background-color)' }],
   ['bg-subtle', { 'background-color': 'var(--w-background-color-subtle)' }],
+  ['bg-inherit', { 'background-color': 'inherit' }],
+  ['bg-transparent', { 'background-color': 'transparent' }],
```
magnuh, line 16: In the tokens color palette there is a variable called `transparent`. I am all for removing that variable... but if we should keep it then we should use it here as well I guess? 😄 
For ref: https://github.com/warp-ds/tokens/pull/18/files#r1149240542

PR #78, opened 2023-03-28
feat(internals): add rules for border-top/right/bottom/left-color classes

BalbinaK:
We've had rules for border width and style for different sides (top/right/bottom/left) and now we also have rules for internal classes that target color of those borders.

### DIFF
--- a/src/_rules/internal.js
+++ b/src/_rules/internal.js
@@ -4,7 +4,7 @@ export const internalRules = [
   [/^i-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.warpToken(cssvar) })],
   [/^i-text-(.+)$/, ([, cssvar]) => ({ color: h.warpToken(cssvar) })],
   [/^i-border-(.+)$/, ([, cssvar]) => ({ 'border-color': h.warpToken(cssvar) })],
-  [/^i-border-([rltb])-(.+)$/, ([_, direction, cssvar]) => {
+  [/^i-border-([rltb])-(.+)$/, ([, direction, cssvar]) => {
     if (direction in directionMap && cssvar != null) {
       return directionMap[direction].map(
         (dir) => [`border${dir}-color`, h.warpToken(cssvar)],
--- a/src/_rules/internal.js
+++ b/src/_rules/internal.js
@@ -1,7 +1,14 @@
-import { handler as h } from '#utils';
+import { handler as h, directionMap } from '#utils';
 
 export const internalRules = [
   [/^i-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.warpToken(cssvar) })],
   [/^i-text-(.+)$/, ([, cssvar]) => ({ color: h.warpToken(cssvar) })],
   [/^i-border-(.+)$/, ([, cssvar]) => ({ 'border-color': h.warpToken(cssvar) })],
+  [/^i-border-([rltb])-(.+)$/, ([_, direction, cssvar]) => {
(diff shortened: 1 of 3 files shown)


src/_rules/internal.js
```diff
@@ -1,7 +1,14 @@
-import { handler as h } from '#utils';
+import { handler as h, directionMap } from '#utils';
 
 export const internalRules = [
   [/^i-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.warpToken(cssvar) })],
   [/^i-text-(.+)$/, ([, cssvar]) => ({ color: h.warpToken(cssvar) })],
   [/^i-border-(.+)$/, ([, cssvar]) => ({ 'border-color': h.warpToken(cssvar) })],
+  [/^i-border-([rltb])-(.+)$/, ([_, direction, cssvar]) => {
```
magnuh, line 7: No big thing 😄 but we don't use underscore variables for non-needed matches in the other regex callbacks.

magnuh: Apparently it works, according to the test output... but I can't understand how. 🤷  🙈 😅 😆

BalbinaK: Removed underscore in [20824bf](https://github.com/warp-ds/drive/pull/78/commits/20824bfe66bd615e8dfe99353905f611d9565d89)

As for why the `/^i-border-(.+)$/` rule didn't work for `i-border-l-$color-alert-warning-border`, it was because the second argument of the callback was then`"l-$color-alert-warning-border"`, and that when passed to the `warpToken` handler didn't match `str.match(/^\$\S/)` check. As a result the CSS property was assigned `undefined` as value, which makes unocss not generate any CSS for that selector. Does that make sense? 😄

AnnaRybkina (approved):
Lgtm 👍 Let's 🚢

src/_rules/internal.js
magnuh: Ah ok I think I get it... it DOES match the first regex and tries to create the border-color rule, but since `h.warpToken()` does not return anything the whole rule gets disposed. It then also matches the next rule, which creates the correct rule using the variable. Did I get that correct? 😄  ...Reading your comment again, I see that was about exactly what you said. 😄 Thanks. 🙏

BalbinaK: As far as I checked, the `cssvar` string needs to start with a `$` to be matched with the regex (it's because we say so with `^` at the beginning of the group). See here:
Screenshot 2023-03-28 at 14 37 26

magnuh: Just trying to understand, and I am NOT saying that you should do this (😅)... but the best way would have been to write the first regex as to match anything but`[rltb]-` so that it wouldn't even try to create the rules? Is that correct? 🤷 🙂

BalbinaK: I think I understand now!
I think the reason for this check to be in the `warpToken` handler was that we'd keep the logic for ensuring correct prefixing in one place. So if we were to change how we separate prefixes from actual tokens, we wouldn't need to do it wherever that token could be used. 
There might also come a time when we rewrite how the color tokens are translated into CSS in our preset, so I'm not sure if it's worth rewriting too much at this point. But that topic is still up for discussion at some point in the future ([related potential task](https://trello.com/c/6kBR0vah))

magnuh: No, nothing that needs to be changed. 🙂  This just made me understand more about how this whole thing is working. 😄  Thanks for taking your time explaining. 🙂 🙏

PR #80, opened 2023-03-28
chore: add bg size to animation

AnnaRybkina:


BalbinaK (approved):
Maybe it was discussed elsewhere or doesn't matter at all, but do you know if we should eventually default to rem values here, too?

AnnaRybkina:
> Maybe it was discussed elsewhere or doesn't matter at all, but do you know if we should eventually default to rem values here, too?

We haven't discussed it. We said that we will use the exact same styling we had before for the loading button, for now. And will discuss if this button will be theme:able in the future. Maybe we also have the colors theme:able as well 🤷 But there is no need now at least

PR #81, opened 2023-04-13
fix: remove input related stuff from preflight

AnnaRybkina:
As we are now styling input in component classes we do not need preflight changes for input

Short discussion about it here: https://sch-chat.slack.com/archives/C050N3NQ204/p1681292111049479

### DIFF
--- a/src/_preflights/forms/form-elements.js
+++ b/src/_preflights/forms/form-elements.js
@@ -3,7 +3,7 @@ export const formElements = `
   user-select: none;
 }
 
-.input, .textarea, .radio, .checkbox, .select {
+.radio, .checkbox, .select {
   display: block;
   border-width: 1px;
   /* specific properties are used here so the focus-ring via outline is instant */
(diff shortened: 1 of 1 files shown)


src/_preflights/forms/form-elements.js
```diff
@@ -13,26 +13,21 @@ export const formElements = `
...
+.select:hover {
   border-color: var(--x-form-field-border-hover);
 }
-.input:focus, .textarea:focus, .select:focus {
+.select:focus {
   border-color: var(--x-form-field-focus);
   outline-color: var(--x-form-field-focus);
 }
 
-.input::placeholder {
```
AnnaRybkina, line 31: This is needed to be handled in the component classes and components.

AnnaRybkina: Wonder if the tw-reset for placeholder will override whatever we set in the components 🤔

BalbinaK: I see it was added to component-classes here: https://github.com/warp-ds/component-classes/commit/f1cdd745582733ac13a286a2d6ab310e5f01943c
Do you think we should add opacity: 1 at some point, too?

AnnaRybkina: @Skadefryd What do you think? Shall we get opacity?

BalbinaK (approved):

PR #83, opened 2023-04-18
fix: height and width shouldn't match without a dash

pearofducks:
Rules like `h32` were matching, which makes it so both `h-32` and `h32` would have the same effect - and it would be ideal to have 'one true way' of writing these.

Slightly worse, Fabric has `h1`, `h2`, and `h3` (f.ex), and this matching would cause these to temporarily look very broken as users migrate.


BalbinaK (approved):
Great catch! Thanks 🚀

PR #88, opened 2023-04-26
fix: add arbitrary width support for border

AnnaRybkina:
We need to add support for adding arbitrary values for border width. We need it for setting border-width: 6 for the toggle radios.

### DIFF
--- a/src/_rules/border.js
+++ b/src/_rules/border.js
@@ -24,6 +24,15 @@ export const borders = [
   [/^border()-(.+)$/, handlerBorderStyle, { autocomplete: "(border)-style" }],
   [/^border-([xy])-(.+)$/, handlerBorderStyle],
   [/^border-([rltb])-(.+)$/, handlerBorderStyle],
+  [
+    /^border-([rltb])?-?\[([0-9])\]$/,
+    handlerArbitraryBorderSize,
+    {
+      autocomplete: [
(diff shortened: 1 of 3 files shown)


src/_rules/border.js
```diff
@@ -24,6 +24,15 @@ export const borders = [
   [/^border()-(.+)$/, handlerBorderStyle, { autocomplete: "(border)-style" }],
   [/^border-([xy])-(.+)$/, handlerBorderStyle],
   [/^border-([rltb])-(.+)$/, handlerBorderStyle],
+  [
+    /^border-([rltb])?-?\[([0-9])\]$/,
```
BalbinaK, line 28: Just curious, why did you choose to limit the range to 9?

BalbinaK (approved):

src/_rules/border.js
AnnaRybkina: Hm, good question. It kinda felt natural to me to have this range. Now when I think of it, maybe the arbitrary values should be whatever number? Wdyt?

BalbinaK: I agree!

AnnaRybkina: Allowing now any digit https://github.com/warp-ds/drive/pull/88/commits/3752d88c732ec41d22f22bbe5a23ce8b568e6002

PR #95, opened 2023-05-03
fix: background arbitrary url rule change

AnnaRybkina:
Due to the limitations in Unocss we needed to change the rule for the background url with arbitrary values. Unocss omitted quotes and therefor we couldn't use the rule we've had previously. This one is working and have been tested in the Toggle component in react

More info [here](https://github.com/unocss/unocss/issues/1182)


BalbinaK (approved):
💪

PR #97, opened 2023-05-09
feat(touch-actions): add support for touch actions

CarolineGard:
add support for touch actions

### DIFF
--- a/test/__snapshots__/static.js.snap
+++ b/test/__snapshots__/static.js.snap
@@ -142,7 +142,22 @@ exports[`static rules do static things 1`] = `
 .mix-blend-initial{mix-blend:initial;}
 .mix-blend-revert{mix-blend:revert;}
 .mix-blend-revert-layer{mix-blend:revert-layer;}
-.mix-blend-unset{mix-blend:unset;}"
+.mix-blend-unset{mix-blend:unset;}
+.touch-auto{touch-action:auto;}
+.touch-none{touch-action:none;}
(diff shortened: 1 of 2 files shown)


test/__snapshots__/static.js.snap
```diff
@@ -142,7 +142,22 @@ exports[`static rules do static things 1`] = `
...
+.touch-pan-right{touch-action:pan-right;}
+.touch-pan-up{touch-action:pan-up;}
+.touch-pan-down{touch-action:pan-down;}
+.touch-pinch-zoom{touch-action:pinch-zoom;}
+.touch-manipulation{touch-action:manipulation;}
+.touch-inherit{touch-action:inherit;}
+.touch-initial{touch-action:initial;}
+.touch-revert{touch-action:revert;}
+.touch-revert-layer{touch-action:revert-layer;}
+.touch-unset{touch-action:unset;}"
```
AnnaRybkina, line 160: Just curious, I can't see those [here](https://warp-ds.github.io/css-docs/touch-action)  Do we really need them? I've checked in tw as well and no such classes there either.

AnnaRybkina: image

AnnaRybkina: I don't know. Just feel that if they are not in tailwind I don't know why we should have them

AnnaRybkina (approved):

BalbinaK (approved):

PR #98, opened 2023-05-10
feat(slider): add border shadow for active and hover states

CarolineGard:
This PR includes hard coded classnames for the slide thumb. 
The current solution is not supported by theming, and has the same color for all themes. 

The difference from fabric is that the shadow is gray instead of blue.

### DIFF
--- a/src/_rules/slider.js
+++ b/src/_rules/slider.js
@@ -1,0 +1,15 @@
+// Hardcoded styling for slider border shadow, will be modified and handled with correct color variables at a later stage
+const defaultColor = 'rgba(0, 0, 0, .08)';
+const sliderStates = {
+  active: '0 0 0 8px',
+  hover: '0 0 0 6px',
+};
+
+export const sliderHandleShadow = [
+  [
+    /^slider-handle-shadow-(active|hover)$/,
+    ([, state]) => ({
+      'box-shadow': `${sliderStates[state]} ${defaultColor})`,
+    }),
+  ],
+];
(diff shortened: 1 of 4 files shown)


src/_rules/slider.js
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export const sliderHandleShadow = [
+  [
+    /^slider-handle-shadow-(active|hover)$/,
+    ([, state]) => ({
+      'box-shadow': `${sliderStates[state]} ${defaultColor})`,
+    }),
+  ],
+];
```
BalbinaK, line 15: Is this something we would like to make available to anyone or is it component specific that fits more to what we have in [internals.js](https://github.com/warp-ds/drive/blob/alpha/src/_rules/internal.js)? There's also a shadow file in https://github.com/warp-ds/drive/blob/alpha/src/_rules/shadow.js that maybe would be a good place for this rule? 

I'm wondering if this box-shadow value can change for different brands, because then we might need to add it to shadow token definitions in @warp-ds/tokens and apply the generated css variable in drive in a similar manner as other shadows.

CarolineGard: yeah I was thinking about those things too but I went with this solution even though. My thoughts:

I would say this is styling specifically for the `slider`. Maybe there will be use cases we don't know of in the future but I would say we can change that later in that case. Also, this is very hardcoded at the moment to make the slider work, so we need to look at this in a later stage and maybe we'll move it then if we have another solution. 

And even though it's box-shadow, I don't think we should put it with the shadows as it's more of a border with a workaround using `box-shadow`. 

what do you think?

BalbinaK: I see. Then maybe we could keep a pattern of adding `Todo` in the comments so it's easier to look them up when the time comes, just in case we miss this :)

BalbinaK (approved):

BalbinaK (approved):
💪

PR #99, opened 2023-05-10
feat(rules): add typography rules

BalbinaK:
## Description
This PR adds rules for typography-related classes that rely on font-size and line-height tokens coming from the theme css file (or preflight css):
- add rules for `text-{number}`, `text-{t-shirt-size}`, `leading-{number}`, `leading-{t-shirt-size}`
- add aliases for the above classes, like `h1`, `t1`, `text-body` using [unocss shortcuts](https://github.com/unocss/unocss/blob/main/docs/config/shortcuts.md)
- define default values for font-size and line-height CSS variables in Preflight, so it's possible to use the above rules without setting theme-specific typography tokens

Before:
Screenshot 2023-05-10 at 15 18 31

After:
Screenshot 2023-05-10 at 15 00 44

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "#plugin": "./src/plugin.js",
     "#utils": "./src/_utils/index.js",
     "#rules": "./src/_rules/index.js",
+    "#shortcuts": "./src/_shortcuts/index.js",
     "#theme": "./src/theme.js",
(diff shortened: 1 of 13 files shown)


BalbinaK:
[Add typography tokens](https://trello.com/c/ONEZZuP7/213-add-typography-tokens)

PR #101, opened 2023-05-17
fix: add base styles to preflight

AnnaRybkina:
We need some styles for the text and anchors. Those are taken from 👇 
https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/base.css#L19
and 👇 
https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/base.css#L47

### DIFF
--- a/src/_preflights/tw-reset.js
+++ b/src/_preflights/tw-reset.js
@@ -35,12 +35,14 @@ html {
 
 /*
 1. Remove the margin in all browsers.
-2. Inherit line-height from 'html' so users can set them as a class directly on the 'html' element.
+2. Base font size.
+3. Base line height.
 */
 
 body {
   margin: 0; /* 1 */
-  line-height: inherit; /* 2 */
+  font-size:var(--w-font-size-m); /* 2 */ 
+  line-height:var(--w-line-height-m); /* 3 */
 }
 
 /*
@@ -82,8 +84,13 @@ Reset links to optimize for opt-in styling instead of opt-out.
 */
 
 a {
-  color: inherit;
+  color: var(--w-color-text-link);
   text-decoration: inherit;
+}
+
+a:hover {
+  color: var(--w-color-text-link);
+  text-decoration: underline;
 }
 
 /*


src/_preflights/tw-reset.js
```diff
@@ -82,10 +84,15 @@ Reset links to optimize for opt-in styling instead of opt-out.
 */
 
 a {
-  color: inherit;
+  color: var(--w-color-text-link);
   text-decoration: inherit;
 }
 
+a:hover {
```
BalbinaK, line 91: Why was a:active and a:focus omitted here if we were applying styles from [Fabric's base.css](https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/base.css#LL53C1-L55C8)?

src/_preflights/tw-reset.js
```diff
@@ -82,10 +84,15 @@ Reset links to optimize for opt-in styling instead of opt-out.
 */
 
 a {
-  color: inherit;
+  color: var(--w-color-text-link);
   text-decoration: inherit;
 }
```
BalbinaK, line 89: Should we also apply `cursor-pointer` and `no-underline` classes here as in[ Fabric's base.css](https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/base.css#L49-L50)?

src/_preflights/tw-reset.js
```diff
@@ -35,12 +35,14 @@ html {
...
+2. Base font size.
+3. Base line height.
 */
 
 body {
   margin: 0; /* 1 */
-  line-height: inherit; /* 2 */
+  font-size:var(--w-font-size-m); /* 2 */ 
+  line-height:var(--w-line-height-m); /* 3 */
 }
```
BalbinaK, line 46: In the [Typography task](https://trello.com/c/ONEZZuP7/213-add-typography-tokens) we were supposed to style the `body` with the following:
```
body {
    /* Set the default site wide bg-color, text-colour, and default font-size with line-height */
    @apply bg text text-m;
   }
```
I haven't come that far but it looks like you covered a similar issue here 😊

BalbinaK (commented):
Awesome that you pointed out that we haven't included [Fabric's base.css](https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/base.css) resets in Warp. It looks like there are more things that we should update here to cover whatever was there in Fabric (like `ul`, `ol` styling). There are also some other examples of `inherit` or `line-height` values in `tw-reset.js` that we probably shouldn't use. @CarolineGard I think we need to add a task to go through the above-mentioned `base.cs` file and make respective updates in the `tw-reset.js`. Otherwise things will not look as they should. @Skadefryd might have some more insight into why this is necessary.

src/_preflights/tw-reset.js
AnnaRybkina: I don't actually remember why? 🤔 I'll check

AnnaRybkina: We added everything 😁

BalbinaK (approved):

PR #102, opened 2023-05-17
feat(slider): add outline-none styling for slider

CarolineGard:
Styling specific for the slider. 
The outline styling is needed to make the focus state behave according to the design (the current design adds shadow and border)

### DIFF
--- a/src/_rules/slider.js
+++ b/src/_rules/slider.js
@@ -14,3 +14,13 @@ export const sliderHandleShadow = [
     }),
   ],
 ];
+
+export const sliderOutlineNone = [
+  [
+    /^slider-handle-outline-none$/,
+    () => ({
+      outline: 'rgba(0, 0, 0, 0) solid 2px',
+      'outline-offset': '2px',
+    }),
+  ],
+];
--- a/test/__snapshots__/slider.js.snap
+++ b/test/__snapshots__/slider.js.snap
@@ -9,3 +9,13 @@ exports[`slider > get hover slider shadow 1`] = `
 "/* layer: default */
 .slider-handle-shadow-hover{box-shadow:0 0 0 6px rgba(0, 0, 0, .08);}"
 `;
+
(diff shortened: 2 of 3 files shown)


src/_rules/slider.js
```diff
@@ -14,3 +14,13 @@ export const sliderHandleShadow = [
     }),
   ],
 ];
+
+export const sliderOutlineNone = [
+  [
+    /^slider-handle-outline-none$/,
+    () => ({
+      outline: 'rgba(0, 0, 0, 0) solid 2px',
+      'outline-offset': '2px',
```
BalbinaK, line 23: I remember that there was an `outline-none` class before that would apply the `outline` and `outline-offset` as you did here, but it was said to be deprecated. Is this a use-case where we'd consider bringing it back? @Skadefryd @adidick @AnnaRybkina?

BalbinaK (commented):
> Styling specific for the slider. The outline styling is needed to make the focus state behave according to the design (the current design adds shadow and border)

Which design are you referring to? Would be really helpful with some before-after screenshots if possible, just to make the review process a bit easier 🙏

src/_rules/slider.js
AnnaRybkina: We do have https://github.com/warp-ds/drive/blob/alpha/src/_rules/focus-ring.js#L13 that adds `outline-none` but it is totally different I guess.

AnnaRybkina: I would suggest to add a very clear TODO here(if we are going with this solution) that this way is not a preferred way of adding styling.
@Skadefryd is also suggesting that we add `outline-offset` rule to drive to handle line 23 as it is done in [TW](https://tailwindcss.com/docs/outline-offset)

CarolineGard: Okay agree, but we can go with this solution for now and add a todo, as it's a copy from the fabric solution? And we can modify it later when outline-offset is in place?

PR #103, opened 2023-05-19
feat: rule for adding safe area inset from the bottom edge of the viewport

AnnaRybkina:
While migrating styles for the modal the need of adding the safe inset appeared. 
Here are two code examples 👇 
https://github.com/fabric-ds/css/blob/next/src/components/modal.css#L16
https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/components/toast.css#L3

I couldn't find any suitable rule in unocss, only custom plugins. So I ended up creating our own rules. For the naming I got inspired by [this plugin](https://github.com/mvllow/tailwindcss-safe-area)

Wdyt?

### DIFF
--- a/src/_rules/static.js
+++ b/src/_rules/static.js
@@ -189,6 +189,7 @@ export const touchAction = [
 ];
 
 export const safeArea = [
-  ['pb-safe', { 'padding-bottom': 'calc(32px + env(safe-area-inset-bottom, 0px))' }],
+  ['pb-safe', { 'padding-bottom': 'env(safe-area-inset-bottom, 0px)' }],
   ['mb-safe', { 'margin-bottom': 'env(safe-area-inset-bottom, 0px)' }],
+  [/^pb-safe-\[([\d]+)]$/, ([, d]) =>  ({ 'padding-bottom': `calc(${d}px + env(safe-area-inset-bottom, 0px))` })],
 ];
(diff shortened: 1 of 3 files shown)


src/_rules/static.js
```diff
@@ -187,3 +187,8 @@ export const touchAction = [
   ['touch-pinch-zoom', { 'touch-action': 'pinch-zoom' }],
   ['touch-manipulation', { 'touch-action': 'manipulation' }],
 ];
+
+export const safeArea = [
+  ['pb-safe', { 'padding-bottom': 'calc(32px + env(safe-area-inset-bottom, 0px))' }],
```
AnnaRybkina, line 192: This `32px` is a bit custom and we maybe can do some smartness there but I can't come up with anything quick here 🤔 Ideas? Or just leave it for now?

BalbinaK: I guess it's fine to leave it like this for now. 
But just from the top of my head - maybe in the future we could allow an arbitrary value that would be used to calculate the safe area inset, if provided? Then it would work like this:
- `pb-safe` -> `{ 'padding-bottom': 'env(safe-area-inset-bottom, 0px)' }`
- `pb-safe-[32]` -> `{ 'padding-bottom': 'calc(32px + env(safe-area-inset-bottom, 0px))' }`

BalbinaK (approved):

src/_rules/static.js
AnnaRybkina: That is great suggestion. I will implement it directly 💪

AnnaRybkina: https://github.com/warp-ds/drive/pull/103/commits/9eb6e7ae9f86451f349be0473f2ea96f6c173191

BalbinaK (approved):

PR #105, opened 2023-05-23
feat(dropshadow): add drop shadow as a constant for use in attention

CarolineGard:
This PR includes a constant fix for `drop shadows`. 
the `filter: drop-shadow ...` is needed for the `attention popover` component

This will later be modified if we need it in other places. Should in that case later support dynamic rgba values, as for `shadows`

### DIFF
--- a/src/_rules/shadow.js
+++ b/src/_rules/shadow.js
@@ -10,3 +10,8 @@ const boxShadowDefaults = {
 export const shadows = [
   [/^shadow-(small|medium|large|xlarge)$/, ([, size]) => ({ 'box-shadow': `var(--w-shadow-${size}, ${boxShadowDefaults[size]})` })],
 ];
+
+
+export const dropShadow = [
+  [/^drop-shadow$/, () => ({ 'filter': 'drop-shadow(rgba(64, 64, 64, 0.24) 0px 3px 8px) drop-shadow(rgba(64, 64, 64, 0.16) 0px 3px 6px)' })],
+];
--- a/test/shadow.js
+++ b/test/shadow.js
@@ -8,3 +8,9 @@ test('shadows work', async ({ uno }) => {
   const { css } = await uno.generate(classes);
   expect(css).toMatchSnapshot();
 });
+
+test('drop shadows', async ({ uno }) => {
+  const classes = ['drop-shadow'];
+  const { css } = await uno.generate(classes);
+  expect(css).toMatchSnapshot();
+});
--- a/test/__snapshots__/shadow.js.snap
+++ b/test/__snapshots__/shadow.js.snap
@@ -1,5 +1,20 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`drop shadow s 1`] = `
+"/* layer: default */
+.drop-shadow{filter:drop-shadow(rgba(64, 64, 64, 0.24) 0px 3px 8px) drop-shadow(rgba(64, 64, 64, 0.16) 0px 3px 6px);}"
+`;
+
+exports[`drop shadow work 1`] = `
+"/* layer: default */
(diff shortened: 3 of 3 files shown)


src/_rules/shadow.js
```diff
@@ -9,4 +11,17 @@ const boxShadowDefaults = {
...
\ No newline at end of file
+];
+
+const dropShadowDefaultColor = '64, 64, 64';
+
+const dropShadowDefaults = {
+  small: [`drop-shadow(rgba(${dropShadowDefaultColor}, .16) 0 1px 6px) drop-shadow(rgba(${dropShadowDefaultColor}, .24) 0 1px 1px)`],
+  medium: [`drop-shadow(rgba(${dropShadowDefaultColor}, .24) 0 3px 8px) drop-shadow(rgba(${dropShadowDefaultColor}, .16) 0 3px 6px)`],
+  large: [`drop-shadow(rgba(${dropShadowDefaultColor}, .23) 0 6px 8px) drop-shadow(rgba(${dropShadowDefaultColor}, .19) 0 10px 20px)`],
+  xlarge: [`drop-shadow(rgba(${dropShadowDefaultColor}, .22) 0 9px 12px) drop-shadow(rgba(${dropShadowDefaultColor}, .25) 0 14px 28px)`],
```
BalbinaK, line 22: Should we maybe rename these to -s, -m, -l and -xl as discussed at standup? Would be great if you could also add a todo right next to box-shadow rules to do the same when we have time (or add a task wherever it makes sense). Wdyt?

test/shadow.js
```diff
@@ -8,3 +8,9 @@ test('shadows work', async ({ uno }) => {
   const { css } = await uno.generate(classes);
   expect(css).toMatchSnapshot();
 });
+
+test('drop shadows', async ({ uno }) => {
+  const classes = ['drop-shadow-small', 'drop-shadow-medium', 'drop-shadow-large', 'drop-shadow-xlarge'];
```
BalbinaK, line 13: this should probably be updated too :)

src/_rules/shadow.js
```diff
@@ -7,6 +9,20 @@ const boxShadowDefaults = {
...
+
+const dropShadowDefaults = {
+  s: [`drop-shadow(rgba(${dropShadowDefaultColor}, .16) 0 1px 6px) drop-shadow(rgba(${dropShadowDefaultColor}, .24) 0 1px 1px)`],
+  m: [`drop-shadow(rgba(${dropShadowDefaultColor}, .24) 0 3px 8px) drop-shadow(rgba(${dropShadowDefaultColor}, .16) 0 3px 6px)`],
+  l: [`drop-shadow(rgba(${dropShadowDefaultColor}, .23) 0 6px 8px) drop-shadow(rgba(${dropShadowDefaultColor}, .19) 0 10px 20px)`],
+  xl: [`drop-shadow(rgba(${dropShadowDefaultColor}, .22) 0 9px 12px) drop-shadow(rgba(${dropShadowDefaultColor}, .25) 0 14px 28px)`],
+};
+
+export const dropShadows = [
+  [/^drop-shadow-(s|m|l|xl)$/, ([, size]) => ({ 'filter': dropShadowDefaults[size] })],
```
BalbinaK, line 27: In the case od box-shadow we used the shadow tokens and applied defaults if those were missing. Here I see we set defaults only. I asked if that also should be the case here in https://github.com/warp-ds/component-classes/pull/57#discussion_r1202603931.

CarolineGard: What do you mean? Is this a blocker or can we iterate on this later?

Skadefryd (approved):
We can always add the tokenstuff later if we see we need them :)

src/_rules/shadow.js
```diff
@@ -7,6 +9,20 @@ const boxShadowDefaults = {
   xlarge: `0 9px 12px ${defaultColor}, 0 14px 28px ${defaultColor}`,
 };
 
+// TODO: rename to -s, -m, -l, -xs
```
AnnaRybkina, line 12: Isn't it done now?

PR #106, opened 2023-05-24
fix: remove resets from preflight and refactor

AnnaRybkina:
As we no longer need resets(the code for it) here (added here instead https://github.com/warp-ds/css) we need to do a refactoring of the preflight related stuff here.

Introduced `skipPreflight` as it is now the option for opting out from `preflight`. Before we needed to `opt in` but now it is another way around.

`development` flag is removed as well. Changed will be needed to be done in the repos using it, like vue


Skadefryd (approved):

src/plugin.js
```diff
@@ -9,15 +9,14 @@ import { postprocess } from '#postprocess';
  * @typedef PluginOptions
  * @type {Object}
  * @property {boolean} development // not in use yet
- * @property {boolean} usePreflight - force preflights to be included/excluded
+ * @property {boolean} test // internal use only
```
BalbinaK, line 12: I'm wondering how intuitive it will be to set a `test` flag compared to something like `skipPreflight` or `noPreflight`. `test` flag does basically one thing here - remove preflights from the plugin. If you think it may do more than that then ignore this comment 😄

src/_preflights/resets.js
```diff
@@ -0,0 +1,4 @@
+export const resets = {
+  layer: 'preflights',
+  getCSS: async () => (await fetch('https://assets.finn.no/pkg/@warp-ds/css/v1/resets.css')).text(),
```
BalbinaK, line 3: We'd probably need to update that repo with changes from e.g. https://github.com/warp-ds/drive/pull/107/files

AnnaRybkina: Yes, absolutely. Just waiting for us to land on something here.

src/plugin.js
AnnaRybkina: Yes, for sure. Name could be improved. Your suggestions sound good 👍🏻

src/_preflights/resets.js
AnnaRybkina: Done

src/plugin.js
AnnaRybkina: [Added](https://github.com/warp-ds/drive/pull/106/commits/f615e6eafbbd1cd03c6fb9e850d1c0a3dd82b197) `skipPreflight` instead and refactored.

BalbinaK (approved):
🎉

src/_preflights/resets.js
BalbinaK, line 3: I just noticed that we also make a minified version available on eik ([@warp-ds/css PR](https://github.com/warp-ds/css/pull/2/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R11-R17)). There it was mentioned that based on whether it's used in production or in development, the minified or non-minified version can be used. Do you think we can keep the non-minified version here or should we rather replace it with the minified since we're not really saying anywhere that we're in development mode? Just checking, I don't mind either option.

AnnaRybkina: Exactly, I've added to have discussion about it. Don't think I added any comment about the discussion though 🤦 Sorry. I **_think_** that we should have minified version here. And then fix so we can use non minified version in dev. But then maybe we will need a dev flag for it? 🤔 
Maybe the non minified step is just too much. What do you think?

BalbinaK: If we do keep dev flag, we'd need to find some smart way of using that to conditionally return minified or non-minified version of the resets.css. The only solution I could think of from the top of my head would be to export two constants from this file, `minResets` and `resets`, and then apply one of them based on `options.development`. But I'm sure there are smarter ways 😄

AnnaRybkina: Shall we do it you think? I'm starting leaning towards only having the minified version 😅 And if we will need the unminified one then we can add it. Preflights are very static and won't change a lot I suppose so maybe there is no point to complicate the setup for that 🤔 wdyt?

BalbinaK: I totally agree! Let's keep the minified version only until suggested otherwise 👍

AnnaRybkina: Changed here https://github.com/warp-ds/drive/pull/106/commits/274a6f7cd8f21f21d77baa83171d63698669b83e. Awaiting css repo changes so the minified version can be published

PR #107, opened 2023-05-25
fix: restructure html and body styling to make shadow dom styling work

AnnaRybkina:
Font-scaling #issue
Rem sized fonts always reference whatever is set on the html tag, in most cases the html will have the same warp reset and will work fine, but what do we do in cases where we are trying to run a warp thing in shadow dom on templates that are NOT warp ?   

If template runs Fabric:   
In theory we could release fabric with a patch that just sets the html reference to 62.5% and it shouldn’t affect the rest of the site since fabric is set up with locked down px sizes. (A quick test on finn.no front-page showed one single text affected by adding that)

On Other sites:
We need to look into how these works case by case. 
If Rems are in use for font-scaling and is scaling differently than 62.5% then it will prob be a world of pain.

This fix in this PR addresses problem with missing body styling in :host. Restructure a bit. Retested. in Vue(no shadow dom) and Tech-docs(with shadow dom + font-size setting in html) all works

### DIFF
--- a/src/_preflights/tw-reset.js
+++ b/src/_preflights/tw-reset.js
@@ -17,37 +17,39 @@ export const twReset = {
 
 /*
 1. Base font size.
-2. Prevent adjustments of font size after orientation changes in iOS.
-3-4. Use a more readable tab size.
-5. Use transparent color of the highlight that appears over an element while it's being tapped
 */
 
-html, :host {
+html {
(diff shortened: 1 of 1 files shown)


Skadefryd (approved):
Hmmm i dont know! , Ok then!

PR #108, opened 2023-05-26
fix: improve syntax to clear the float for the elements after legend

AnnaRybkina:
I could reproduce issue in Firefox [here](https://warp-ds.github.io/tech-docs/components/checkbox/) and then in local Vue setup
Expected|What we got|Fixed
---|---|---
image|image|image


magnuh (approved):

CarolineGard (approved):
Nice!

PR #109, opened 2023-05-29
fix: focusable classes

AnnaRybkina:
Migrated fabric stuff from https://github.com/warp-ds/drive/blob/alpha/src/_rules/focus-ring.js but didn't work as we wished.

Before|After
---|---
image|image

### DIFF
--- a/src/_rules/focus-ring.js
+++ b/src/_rules/focus-ring.js
@@ -1,16 +1,24 @@
+import { entriesToCss } from '@unocss/core';
+
 // TODO: use actual variables and values when those has been defined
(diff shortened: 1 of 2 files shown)


src/_rules/focus-ring.js
```diff
@@ -1,16 +1,24 @@
+import { entriesToCss } from '@unocss/core';
+
 // TODO: use actual variables and values when those has been defined
```
magnuh, line 3: Is this an old comment? 🤔

magnuh (approved):

PR #119, opened 2023-06-12
feat: Add background, text and border classes for semantic color tokens

magnuh:
The idea is to give consumers access to semantic variables to set background, border and text colors.

Using the class `s-bg-default` would give the following css: `background-color: var(--w-s-color-background-default);`

`s-border-positive-default` would give: `border-color: var(--w-s-color-border-positive-default);`
(also including support for directions [lrtbxy], e.g. `s-border-x-positive-default`)

and `s-text-link-hover` would give: `color: var(--w-s-color-text-link-hover);`

...based on the tokens added in this PR: https://github.com/warp-ds/tokens/pull/73

### DIFF
--- a/src/_rules/semantic.js
+++ b/src/_rules/semantic.js
@@ -1,0 +1,7 @@
+import { handler as h } from '#utils';
+
+export const semanticRules = [
+  [/^s-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.semanticToken(`background-${cssvar}`) })],
+  [/^s-text-(.+)$/, ([, cssvar]) => ({ color: h.semanticToken(`text-${cssvar}`) })],
+  [/^s-border-(.+)$/, ([, cssvar]) => ({ 'border-color': h.semanticToken(`border-${cssvar}`) })],
+];
--- a/src/_rules/semantic.js
+++ b/src/_rules/semantic.js
@@ -1,7 +1,10 @@
-import { handler as h } from '#utils';
+import { directionMap, handler as h } from '#utils';
 
 export const semanticRules = [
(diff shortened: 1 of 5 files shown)


CarolineGard (commented):
Cool, looking good 👏

src/_rules/semantic.js
```diff
@@ -0,0 +1,7 @@
+import { handler as h } from '#utils';
+
+export const semanticRules = [
+  [/^s-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.semanticToken(`background-${cssvar}`) })],
+  [/^s-text-(.+)$/, ([, cssvar]) => ({ color: h.semanticToken(`text-${cssvar}`) })],
```
BalbinaK, line 5: I saw icon tokens, too. Should they also be added here?

AnnaRybkina (changes requested):
Looks great 🙌 Nice job. I miss tests though, I think it is a good practice to have snapshots so we have a safety level if something changes or if something happens to collide with something else. Wdyt?

src/_rules/semantic.js
magnuh: Yes, probably... had a question about that in the slack dev-channel. 😄

magnuh:
> Looks great 🙌 Nice job. I miss tests though, I think it is a good practice to have snapshots so we have a safety level if something changes or if something happens to collide with something else. Wdyt?

Yes, already informally requested, so I'm working on it. 😄 

Also there need to be directions for the border classes... coming up.

AnnaRybkina (approved):

BalbinaK:
[Add semantic tokens](https://trello.com/c/2IPXw986/253-add-semantic-tokens)

Issue #117, opened 2023-06-08
Border


Skadefryd:
border-x, and border-y not working.  

Ment to do the same as border-x-1, and border-y-1, but should be there for consistency. 

https://warp-ds.github.io/css-docs/border-width

BalbinaK:
@AnnaRybkina do you know if this issue can be closed?

AnnaRybkina:
> @AnnaRybkina do you know if this issue can be closed?

I think we need to add those new to the docs and then we can close it. I'll add them now

PR #120, opened 2023-06-13
fix: border-x|y fix and slider obsolete snapshot fix

AnnaRybkina:
Fixing this https://github.com/warp-ds/drive/issues/117

### DIFF
--- a/src/_rules/border.js
+++ b/src/_rules/border.js
@@ -18,7 +18,8 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border()-(\d+)$/, handlerBorder, { autocomplete: "(border)-<directions>" }],
-  [/^border-([xy])-?(\d+)?$/, handlerBorder],
+  [/^border-([xy])$/, handlerBorder],
+  [/^border-([xy])-(\d+)$/, handlerBorder],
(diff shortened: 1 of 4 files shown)


BalbinaK (approved):

src/_rules/border.js
```diff
@@ -18,7 +18,7 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border()-(\d+)$/, handlerBorder, { autocomplete: "(border)-<directions>" }],
-  [/^border-([xy])-(\d+)$/, handlerBorder],
+  [/^border-([xy])-?(\d+)?$/, handlerBorder],
```
magnuh, line 21: Hmm... what will that first question mark do? Match zero or one "-"? 🤔  And the second one is to make the capture group optional...  meaning this will make it possible to have only `border-x` and `border-y`?

magnuh: But it would also accept `border-x12` ? 🤔

AnnaRybkina: Refactored [here](https://github.com/warp-ds/drive/pull/120/commits/942d1c19fe64f53e5323715b278911ad053b97ac)

PR #124, opened 2023-06-15
feat(semantic.js): add rules for semantic icon colors

BalbinaK:
Rules are based on our semantic color tokens from [@warp-ds/tokens](https://github.com/warp-ds/tokens/blob/alpha/tokens/finn.no/semantic.yml#L120-L138)


magnuh (approved):

BalbinaK:
[Add semantic tokens](https://trello.com/c/2IPXw986/253-add-semantic-tokens)

PR #125, opened 2023-06-16
feat: add rules for outline

AnnaRybkina:
New rules for outline based on:
- https://tailwindcss.com/docs/outline-width
- https://tailwindcss.com/docs/outline-style
- https://tailwindcss.com/docs/outline-offset

With `outline-none` now added we can remove https://github.com/warp-ds/drive/blob/alpha/src/_rules/slider.js#L20. Will create a ticket for it

TBD: I'll update the css docs with this new rules

### DIFF
--- a/src/_rules/outline.js
+++ b/src/_rules/outline.js
@@ -6,15 +6,15 @@ const outlineNone = {
 };
 
 export const outlineStyle = [
-  ["outline-none", { ... outlineNone }],
+  ["outline-none", { ...outlineNone }],
   ["outline", { "outline-style": "solid" }],
   ["outline-dashed", { "outline-style": "dashed" }],
   ["outline-dotted", { "outline-style": "dotted" }],
   ["outline-double", { "outline-style": "double" }],
 ];
 export const outlineWidth = [
-  [/^outline-(\d+)$/, ([, w]) => ({ ['outline-width']: lineWidth?.[w ?? 1] }), { autocomplete: `(outline)-(${Object.keys(lineWidth).join('|')})` }],
+  [/^outline-(\d+)$/, ([, w]) => ({ ['outline-width']: lineWidth?.[w ?? 1] }), { autocomplete: `outline-(${Object.keys(lineWidth).join('|')})` }],
 ];
 export const outlineOffset = [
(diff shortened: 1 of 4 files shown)


src/_rules/outline.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const outlineStyle = [
+  ["outline-none", { ... outlineNone }],
+  ["outline", { "outline-style": "solid" }],
+  ["outline-dashed", { "outline-style": "dashed" }],
+  ["outline-dotted", { "outline-style": "dotted" }],
+  ["outline-double", { "outline-style": "double" }],
+];
+export const outlineWidth = [
+  [/^outline-(\d+)$/, ([, w]) => ({ ['outline-width']: lineWidth?.[w ?? 1] }), { autocomplete: `(outline)-(${Object.keys(lineWidth).join('|')})` }],
```
BalbinaK, line 16: I think outline doesn't need to be in `()` in the autocomplete section because it's just one string, no other options within that group to match. Docs on unocss' autocomplete: https://github.com/unocss/unocss/blob/main/docs/tools/autocomplete.md

src/_rules/outline.js
```diff
@@ -0,0 +1,20 @@
+import { lineWidth } from '#theme';
+
+const outlineNone = {
+  outline: '2px solid transparent',
+  'outline-offset': '2px',
+};
+
+export const outlineStyle = [
+  ["outline-none", { ... outlineNone }],
```
BalbinaK, line 9: Apparently it works with as many spaces as we'd like here, but maybe we could change it to `...outlineNone` 😄

BalbinaK:
@CarolineGard do these changes fix this comment by any chance? https://github.com/warp-ds/drive/blob/alpha/src/_rules/slider.js#L18-L28

AnnaRybkina:
> @CarolineGard do these changes fix this comment by any chance? https://github.com/warp-ds/drive/blob/alpha/src/_rules/slider.js#L18-L28

Yes, I mentioned it in the description 😊

src/_rules/outline.js
AnnaRybkina: https://github.com/warp-ds/drive/pull/125/commits/57b14f9397bc6a89446c78354aad71d8530c0009

src/_rules/outline.js
AnnaRybkina: https://github.com/warp-ds/drive/pull/125/commits/57b14f9397bc6a89446c78354aad71d8530c0009

BalbinaK (approved):

PR #126, opened 2023-06-16
fix: proper handle bg arbitrary url as var

AnnaRybkina:
[This refactoring](https://github.com/warp-ds/drive/pull/111) broke the check mark icon for the checkboxes. We needed one more check for the vars as urls.

Before|After
---|---
image|image

This PR is needed to make everything work https://github.com/warp-ds/component-classes/pull/89. I tested locally, link local builds for drive and cc in vue

### DIFF
--- a/src/_rules/background.js
+++ b/src/_rules/background.js
@@ -40,7 +40,9 @@ export const backgrounds = [
   //arbitrary
   [/^bg-\[(.+)\]/, ([, p]) => {
     if (p.startsWith('url')) {
-      return { 'background-image': p };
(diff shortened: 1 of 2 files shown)


AnnaRybkina:
[No check mark icon in the checkbox on hover. Should be a white checkmark there](https://trello.com/c/VjKl2Xhe/305-no-check-mark-icon-in-the-checkbox-on-hover-should-be-a-white-checkmark-there)

BalbinaK (approved):
💪

PR #130, opened 2023-06-22
feat: arbitrary values accepted for leading

AnnaRybkina:
Almost all buttons are using `leading-24` . The class is also used in one place in finn. 24 - is not a part of our tokens so we need to support arbitrary values here. (more info about the discussion [here](https://sch-chat.slack.com/archives/C050N3NQ204/p1686921532451109))

We have also one usage of the custom line-height https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/form-elements.css#L176 that we will be able to set with this support.


CarolineGard (approved):
Nice 🙌

PR #132, opened 2023-06-26
fix: throw better errors for users who are missing 'fetch'

pearofducks:


AnnaRybkina (approved):
Great 👍🏻 Let’s try it out!

PR #133, opened 2023-06-27
fix: Updated rules for semantic colors

magnuh:
Updated rules for semantic colors to handle base level token names without ending "default" suffix.


src/_rules/semantic.js
```diff
@@ -1,10 +1,17 @@
...
 export const semanticRules = [
+  [/^s-bg$/, () => ({ 'background-color': h.semanticToken('background') })],
   [/^s-bg-(.+)$/, ([, cssvar]) => ({ 'background-color': h.semanticToken(`background-${cssvar}`) })],
+  [/^s-text$/, () => ({ color: h.semanticToken('text') })],
   [/^s-text-(.+)$/, ([, cssvar]) => ({ color: h.semanticToken(`text-${cssvar}`) })],
+  [/^s-icon$/, () => ({ color: h.semanticToken('icon') })],
   [/^s-icon-(.+)$/, ([, cssvar]) => ({ color: h.semanticToken(`icon-${cssvar}`) })],
+  [/^s-border$/, ([, cssvar]) => ({ 'border-color': h.semanticToken('border') })],
   [/^s-border-(.+)$/, ([, cssvar]) => ({ 'border-color': h.semanticToken(`border-${cssvar}`) })],
+  [/^s-border-([lrtbxy])$/, ([, direction]) => directionMap[direction]?.map(
```
BalbinaK, line 12: Maybe we could add tests for these rules, too?

CarolineGard (approved):
Nice ⭐

src/_rules/semantic.js
magnuh: Ah, you mean for the color-less ones? 😄

BalbinaK: I thought `s-border-l`, `s-border-y` etc. will use a color token, like `.s-border-l {border-left-color: var(--w-s-border);}`. So I was thinking to add those types of classes to the tests because we only do so with `.s-border-t-positive`, `s-border` and so on. Am I missing something? :D

magnuh: Nope, you are correct. With "color-less" I meant without a color specified in the class name. 😄  I'll add it to the test.

magnuh: Added to the test now. 👍

BalbinaK (approved):

PR #134, opened 2023-06-29
fix: cache the preflight instead of requesting every rebuild

pearofducks:
The `getCSS` function runs every time the plugin is doing a build - which when developing locally is _often_.

Caching the CSS here means our servers will serve less data, and also the local development experience will be sped up by however long this download would have taken; meaning if this download would ever experience issues the user's dev-env would hang for unknown reasons to them!


siljec (approved):
LGTM 🎉

PR #135, opened 2023-06-29
fix: add missing classes Fabric supported

siljec:
[WARP-121](https://nmp-jira.atlassian.net/browse/WARP-121): Add `tabular-nums`, `lining-nums` and `border-inverted` (to replace `border-white` ) + some font variant numeric classes from tailwind.

Other notes: 
* PR to update parity with these findings: https://github.com/warp-ds/parity/pull/3
* New working document to investigate the classes: https://docs.google.com/spreadsheets/d/1W0n74LSV6IG2YIU1LJNopFNtTkSYLY3rfPa9c1qg9d4/edit#gid=0
* We also had this from earlier: https://docs.google.com/spreadsheets/d/1gk1LhpUimjk9QnZhZyId7c56xk30dbzQxRCzuGLlllY/edit#gid=0

### DIFF
--- a/src/_rules/border.js
+++ b/src/_rules/border.js
@@ -18,7 +18,7 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border-transparent$/, () => ({ 'border-color': 'transparent' })],
-  [/^border-white$/, () => ({ 'border-color': 'var(--w-white)' })],
+  [/^border-inverted$/, () => ({ 'border-color': 'var(--w-s-border-inverted)' })],
   [/^border-inherit$/, () => ({ 'border-color': 'inherit' })],
   [/^border-current$/, () => ({ 'border-color': 'currentColor' })],
   [/^border()-(\d+)$/, handlerBorder, { autocomplete: "(border)-<directions>" }],
--- a/src/_rules/border.js
+++ b/src/_rules/border.js
@@ -18,6 +18,7 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border-transparent$/, () => ({ 'border-color': 'transparent' })],
+  [/^border-white$/, () => ({ 'border-color': 'var(--w-white)' })],
   [/^border-inherit$/, () => ({ 'border-color': 'inherit' })],
   [/^border-current$/, () => ({ 'border-color': 'currentColor' })],
   [/^border()-(\d+)$/, handlerBorder, { autocomplete: "(border)-<directions>" }],
--- a/src/_rules/typography.js
+++ b/src/_rules/typography.js
@@ -11,4 +11,5 @@ export const typography = [
   ],
   [/^leading-\[(.+)(rem|px)?\]/, ([, value, unit], context) => ({ 'line-height': resolveArbitraryValues(value, unit, context) })],
   ['tabular-nums', { 'font-family': 'ui-sans-serif,system-ui,-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,sans-serif', 'font-variant-numeric': 'tabular-nums' }],
+  ['lining-nums', { 'font-variant-numeric': 'lining-nums' }],
 ];
--- a/src/_rules/typography.js
+++ b/src/_rules/typography.js
@@ -10,4 +10,5 @@ export const typography = [
     ({ 'line-height': `var(--w-line-height-${size})` }),
   ],
   [/^leading-\[(.+)(rem|px)?\]/, ([, value, unit], context) => ({ 'line-height': resolveArbitraryValues(value, unit, context) })],
+  ['tabular-nums', { 'font-family': 'ui-sans-serif,system-ui,-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,sans-serif', 'font-variant-numeric': 'tabular-nums' }],
 ];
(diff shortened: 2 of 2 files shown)


src/_rules/typography.js
```diff
@@ -10,4 +10,5 @@ export const typography = [
     ({ 'line-height': `var(--w-line-height-${size})` }),
   ],
   [/^leading-\[(.+)(rem|px)?\]/, ([, value, unit], context) => ({ 'line-height': resolveArbitraryValues(value, unit, context) })],
+  ['tabular-nums', { 'font-family': 'ui-sans-serif,system-ui,-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,sans-serif', 'font-variant-numeric': 'tabular-nums' }],
```
siljec, line 13: Missing class and requested from Dave.
It was in fabric here: https://github.com/fabric-ds/css/blob/01fd80a27461cf0d593911e974853f84963d0681/src/utilities.css#L13

src/_rules/typography.js
```diff
@@ -10,4 +10,6 @@ export const typography = [
...
+  ['lining-nums', { 'font-variant-numeric': 'lining-nums' }],
```
siljec, line 14: Used in Finn, and not in Warp yet. It's from [tailwind](https://tailwindcss.com/docs/font-variant-numeric).

src/_rules/typography.js
AnnaRybkina: I found a TODO in the css-docs related page. https://warp-ds.github.io/css-docs/font-variant-numeric#applying-numeric-variants 👇 
`!TODO: Discuss the support for this in warp. We might to have to theme this somehow, these fonts might look too different to some of the brand fonts and perhaps need different weights to match them better to the brand font.`
Is it something we need to consider?

src/_rules/border.js
```diff
@@ -18,6 +18,7 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border-transparent$/, () => ({ 'border-color': 'transparent' })],
+  [/^border-white$/, () => ({ 'border-color': '#fff' })],
```
siljec, line 21: `border-white`: Used in Finn, and not in Warp yet. It's from [tailwind](https://v1.tailwindcss.com/docs/border-color)

src/_rules/typography.js
siljec: Very good catch 🙏 I added a ticket: https://nmp-jira.atlassian.net/browse/WARP-171

src/_rules/border.js
AnnaRybkina: I think we want to use `var(--w-white)` here instead of a hex value

magnuh: Are we sure that we really want a `border-white` class though? 🤷  It says in the missing classes sheets that it is not used in Finn? And if we want to enable dark-mode, we would probably want whatever element this is set on to have a black border instead? We have `inverted` semantic tokens for background, icon and text... so we might want that for border as well? (ping @adidick)

AnnaRybkina: `border-white` is used a lot in Finn. I found a lot of usages. Good point about the inverted tokens. I guess we should do the same for borders, let's see what Adi answers

siljec: Yes, I searched in the wrong place, so it is used a lot. I'll change it to `var(--w-white)` at least :) Thanks!

magnuh: I really don't think that we should add this. Simply because it will break dark-mode (if that is something we'll be doing?). Or... if we add a brand that has a dark background as default, not having a semantic class will give us no chance to override this for that brand... white is white. 🤷  I talked to Adi about this, and I think we should add a `border-inverted` semantic token... so then we can point any usage of `border-white` to instead use `s-border-inverted`. Wdyt?

magnuh: Semantic tokens for inverted border (`s-border-inverted`) added in this PR : https://github.com/warp-ds/css/pull/22

siljec: Good idea! I changed it now @magnuh!

src/_rules/typography.js
```diff
@@ -10,4 +10,10 @@ export const typography = [
...
+  ['ordinal', { 'font-variant-numeric': 'ordinal' }],
+  ['slashed-zero', { 'font-variant-numeric': 'slashed-zero' }],
+  ['oldstyle-nums', { 'font-variant-numeric': 'oldstyle-nums' }],
+  ['proportional-nums', { 'font-variant-numeric': 'proportional-nums' }],
```
siljec, line 18: As discussed with Snorre last week, I'm adding these classes from tailwind also.

src/_rules/border.js
```diff
@@ -18,6 +18,7 @@ const borderStyles = [
 export const borders = [
   [/^border$/, handlerBorder],
   [/^border-transparent$/, () => ({ 'border-color': 'transparent' })],
+  [/^border-inverted$/, () => ({ 'border-color': 'var(--w-s-border-inverted)' })],
```
Skadefryd, line 21: Isnt the class names the only thing running the s- prefix? (not the vars? )

Skadefryd (approved):

src/_rules/border.js
siljec: Hmm I see a lot og `--w-s-xxx` vars here: https://assets.finn.no/pkg/@warp-ds/tokens/1.0.0-alpha.92/finn-no.css - so I think it should contain -s here?

PR #140, opened 2023-06-30
fix: Remove old fallback values from box shadows

magnuh:
Removing old hard coded fallback values for box-shadows as discussed here: https://sch-chat.slack.com/archives/C050N3NQ204/p1688047940344899


siljec (approved):
🚀

PR #143, opened 2023-07-07
fix: use new css repo for component classes

AnnaRybkina:
Now that we have own new css repo which contains component-classes we need to update dependencies and imports

- [x] bump to the latest version of the css before merge

### DIFF
--- a/dev.js
+++ b/dev.js
@@ -12,12 +12,18 @@ const {
       short: 'c',
     },
     externalClasses: {
-      type: 'boolean',
+      type: 'string',
     },
     externalizeClasses: {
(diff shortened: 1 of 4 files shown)


dev.js
```diff
@@ -12,14 +12,20 @@ const {
       short: 'c',
     },
     externalClasses: {
-      type: 'boolean',
+      type: 'string',
```
siljec, line 15: I see that it is used like this:
` const safeExternalClasses = options.externalClasses || [];`

should it be `string[]`?

siljec (approved):

PR #146, opened 2023-08-14
fix: add font-bold to t1-t6 and use those classes to ref h1-h5

BalbinaK:
Titles in WARP should always have bold font, which has been the case in Figma for T1-T6 styles, but was missed in our uno preset. This PR fixes it and also makes those t1-t6 classes the reference for h1-h5 classes.


AnnaRybkina (approved):
Looks good to me 💪 🚢

PR #147, opened 2023-08-15
fix(display.js): align order of css for display classes with fabric.css

BalbinaK:
The order of CSS for display classes is now aligned with [fabric.css](https://assets.finn.no/pkg/@fabric-ds/css/1.2.0/fabric.min.css) that Warp should be based on. This fixes e.g. issues with `flex` class not overriding any other default display style of a given element, like it used to in Fabric.


siljec (approved):
🎉

PR #148, opened 2023-08-16
fix: update & align unocss dependencies

BalbinaK:
When latest (0.55.0) `unocss` package version was installed, an error was thrown from a lower version of `@unocss/core`:

```
import { escapeSelector, createValueHandler, isString, escapeRegExp, CountableSet, BetterMap } from '@unocss/core';
                                                                     ^^^^^^^^^^^^
SyntaxError: The requested module '@unocss/core' does not provide an export named 'CountableSet'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
```

Aligning the two dependencies in this PR fixes the error but some tests had to be slightly adjusted, too.

By the way, `@warp-ds/css` was updated to `1.0.0-alpha.37` and a deprecated `resets.min.css` file was replaced with `resets.css`.

- [x] test with React components (disabled button seemed to not get the right text colour)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@unocss/core": "^0.55.1",
-    "@unocss/preset-mini": "^0.55.1"
+    "@unocss/preset-mini": "^0.52.7"
   },
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^15.0.1",
--- a/package.json
+++ b/package.json
@@ -37,26 +37,26 @@
   "author": "@djh",
   "license": "Apache-2.0",
   "dependencies": {
-    "@unocss/core": "^0.50.6",
-    "@unocss/preset-mini": "^0.50.6"
+    "@unocss/core": "^0.55.1",
+    "@unocss/preset-mini": "^0.55.1"
   },
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^15.0.1",
     "@semantic-release/changelog": "^6.0.2",
     "@semantic-release/git": "^10.0.1",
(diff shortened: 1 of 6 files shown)


test/general.js
```diff
@@ -31,7 +31,7 @@ test(`autocomplete doesn't throw`, async ({ uno }) => {
 });
 
 test(`the HTML parser is sane`, async ({ uno }) => {
-  const { css } = await uno.generate(`<div class="before:content-['before-stuff'] sm:grid"></div>`);
+  const { css } = await uno.generate([`before:content-['before-stuff']`, 'sm:grid']);
```
BalbinaK, line 34: There was an issue with `"` and `'` in the same string, causing unocss to not generate any css for these classes.

package.json
```diff
@@ -37,26 +37,26 @@
   "author": "@djh",
   "license": "Apache-2.0",
   "dependencies": {
-    "@unocss/core": "^0.50.6",
-    "@unocss/preset-mini": "^0.50.6"
+    "@unocss/core": "^0.55.1",
+    "@unocss/preset-mini": "^0.52.7"
```
BalbinaK, line 41: This dependency has to be at 0.52.7 for now. 
In [0.53.0](https://github.com/unocss/unocss/pull/2714) pseudo variants generation order changes.
That causes changes to css order for classes that control the same CSS properties,
which results in bugs in our components (e.g. Button disabled gets a non-disabled text color).
Before we update this package, we should change the way we assign utility classes to components
when they control the same CSS properties.

siljec: Nice, here's the ticket to do that btw: https://nmp-jira.atlassian.net/browse/WARP-231

siljec (approved):

PR #149, opened 2023-08-17
fix: remove @warp-ds/css dep to fix circular dependency

BalbinaK:
`@warp-ds/css` dependency was only used to import component classes array and “externalize” those to avoid generating redundant CSS when it’s already available in an app through [components.css](https://assets.finn.no/pkg/@warp-ds/css/v1/components.css).
Until we figure out how to fix externalizing of component classes in a better way, we remove the option `omitComponentClasses` from the plugin. The users will then need to import the `@warp-ds/css` package and pass the imported component classes array to `externalClasses` option of `presetWarp` to achieve the same goal.

### DIFF
--- a/test/external-classes.js
+++ b/test/external-classes.js
@@ -9,7 +9,7 @@ describe('external classes', () => {
     const { css } = await uno.generate(classes);
     expect(css).toMatchSnapshot();
   });
-  test('externaliized class of mt-32 should be excluded from the css', async ({ uno }) => {
+  test('externalized class of mt-32 should be excluded from the css', async ({ uno }) => {
     const classes = ['mt-16', 'mt-32', 'mb-24'];
     const { css } = await uno.generate(classes);
     expect(css).toMatchSnapshot();
--- a/test/external-classes.js
+++ b/test/external-classes.js
@@ -4,7 +4,7 @@ import { describe, expect, test } from 'vitest';
 setup({ externalizeClasses: true, externalClasses: ["mt-32", "mb-32"] });
 
 describe('external classes', () => {
-  test('externaliized classes of mt-32 and mb-32 should be excluded from the css', async ({ uno }) => {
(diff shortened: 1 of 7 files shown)


test/external-classes.js
```diff
@@ -0,0 +1,22 @@
+import { setup } from './_helpers.js';
+import { describe, expect, test } from 'vitest';
+
+setup({ externalizeClasses: true, externalClasses: ["mt-32", "mb-32"] });
+
+describe('external classes', () => {
+  test('externaliized classes of mt-32 and mb-32 should be excluded from the css', async ({ uno }) => {
```
AnnaRybkina, line 7: Suuuperminor nitpick
```suggestion
  test('externalized classes of mt-32 and mb-32 should be excluded from the css', async ({ uno }) => {
```

test/external-classes.js
```diff
@@ -0,0 +1,22 @@
...
+    const classes = ['mt-16', 'mt-32', 'mb-32'];
+    const { css } = await uno.generate(classes);
+    expect(css).toMatchSnapshot();
+  });
+  test('externaliized class of mt-32 should be excluded from the css', async ({ uno }) => {
```
AnnaRybkina, line 12:
```suggestion
  test('externalized class of mt-32 should be excluded from the css', async ({ uno }) => {
```

test/external-classes.js
```diff
@@ -0,0 +1,22 @@
...
+    const classes = ['mt-16', 'mt-32', 'mb-24'];
+    const { css } = await uno.generate(classes);
+    expect(css).toMatchSnapshot();
+  });
+  test('externaliized class of mt-32 should be included if used with modifier', async ({ uno }) => {
```
AnnaRybkina, line 17:
```suggestion
  test('externalized class of mt-32 should be included if used with modifier', async ({ uno }) => {
```

PR #156, opened 2023-09-06
feat: Add rules for backdrop-blur

magnuh:
New rules for backdrop-blur based on:
https://tailwindcss.com/docs/backdrop-blur
(though the blur sizes has been modified and also size names are changed to align with how it is in warp)

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -64,8 +64,6 @@ These are TW2 selectors, TW3 and other utility CSS have moved on in syntax.
 
 # Filters
 
-All filters
-
 - filter
 - filter-none
 - blur-sm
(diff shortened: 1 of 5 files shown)


DEPRECATIONS.md
```diff
@@ -120,13 +118,11 @@ All filters
...
-- backdrop-blur-md
-- backdrop-blur-lg
-- backdrop-blur-xl
-- backdrop-blur-2xl
-- backdrop-blur-3xl
+- backdrop-blur-sm :: prefer backdrop-blur-s
+- backdrop-blur-md :: prefer backdrop-blur-m
+- backdrop-blur-lg :: prefer backdrop-blur-l
+- backdrop-blur-2xl :: prefer backdrop-blur-xxl
+- backdrop-blur-3xl :: prefer backdrop-blur-xxxl
```
AnnaRybkina, line 125: Nice 👍 Will we need to update the messages in preset-migrate with these new recommendations

AnnaRybkina (approved):
Looks good. So the default is `4px` for WARP? Haven't checked if that is correct but I guess you know what you are doing 👍

magnuh:
> Looks good. So the default is `4px` for WARP? Haven't checked if that is correct but I guess you know what you are doing 👍

There is nothing specified yet... but it is wanted for the price badges on top of images. And 4px for small as it is in Tailwind looks a bit too much ... so I adjusted the scale.

DEPRECATIONS.md
magnuh: I'll fix.

PR #160, opened 2023-10-04
feat: purge component classes

AnnaRybkina:
A part of https://nmp-jira.atlassian.net/jira/software/projects/WARP/boards/15?label=web&selectedIssue=WARP-302

As we are publishing the component.css to EIK our consumers do not need to safelist the classes anymore. They can use the css directly from EIK. Benefits: cached css instead of every app safelists the classes. Could be added to the template etc

And as we don't want duplicated styles to be generated we can finally use externalizedClasses and postprocess feature. We will ALWAYS purge classes so it is crucial for our consumers to use `warp.css` somewhere in their setup.

- [x] would like to skip fetching if `development: true`
- [x] fix tests

????? consumers will need to add an async wrapper :(
 
In the uno.config 👇 

Before|After
---|---
|<pre><code>export default (() => {</code><br>  <code>return defineConfig({</code><br><code>    presets: [presetWarp()],  </code><br><code>  });</code><br><code>})();</code></pre> | <pre><code>export default (async () => {</code><br><code>  const asyncPreset = await presetWarp()</code><br><code>  return defineConfig({</code><br><code>    presets: [asyncPreset]</code><br><code>  })</code><br><code>})();</code></pre>|

**TO TEST**
This PR is needed to be tested together with https://github.com/warp-ds/css/pull/56
- run build in both packages and link them both in the test application you want to test in, for example any of the component frameworks repos (YOU'LL NEED TO REMOVE SAFELIST FROM THE CONFIG AND ADD `warp.css` where it is applicable)
- observe that everything looks like before
- add e.g. `p-16` classes to any element
- observer `p-16` in the generated uno.css
- change `development` to false and rebuild
- observe that `p-16` is no longer in the uno-css


test/_helpers.js
```diff
@@ -2,17 +2,26 @@ import { createGenerator } from '@unocss/core';
...
+  });
+  return createGenerator({ presets: [asyncPreset] });
+};
 export const setup = (opts = {}) => {
-  beforeEach(t => {
-    t.uno = getGenerator({ ...opts, development: true });
+  beforeEach(async (t) => {
+    t.uno = await getGenerator({
+      ...opts,
+      development: true,
```
BalbinaK, line 16: It was there before but now I wonder if we need to pass `development: true` again if the `getGenerator` already sets that option.

dev.js
```diff
@@ -17,13 +16,20 @@ const {
...
       type: 'boolean',
     },
   },
 });
-
-const uno = createGenerator({ presets: [presetWarp( { ...options, development: true } )] });
+const asyncPreset = await presetWarp({
+  ...options,
+  omitComponentClasses: true,
+  development: true,
```
BalbinaK, line 30: It looks like we intend to omit component classes here. The way I understand the plugin.js logic (see below), the external classes applied with `omitComponentClasses` flag will not be considered in the postprocessor.js. The constant below should be `true` for this to work, but it will be `false` when `options.extenalizeClasses` flag is `undefined` and `options.development` is `true`. For this to work we'd need to pass `externalizeClasses: true`.

```
const externalizeClasses = options.externalizeClasses ?? !options.development;
```

dev.js
```diff
@@ -17,13 +16,20 @@ const {
...
     usePixels: {
       type: 'boolean',
     },
   },
 });
-
-const uno = createGenerator({ presets: [presetWarp( { ...options, development: true } )] });
+const asyncPreset = await presetWarp({
+  ...options,
+  omitComponentClasses: true,
```
BalbinaK, line 29: Do we want to omit component classes when testing what UnoCSS generates in the terminal (`pnpm dev -c someclass`)? Then we won't get any output for classes used in components, which might be confusing when we're checking why some classes allegedly don't work with our preset. Wdyt?

src/classesToPurge.js
```diff
@@ -0,0 +1,17 @@
+let classes;
+async function getClassesToPurge() {
+  if (classes) return classes;
+  else {
+    classes = await fetch(
+      'https://warp-ds.github.io/css/component-classes/classesToPurge.json',
```
BalbinaK, line 6: Was it considered to put this json file to the Eik CDN? Then we'd make sure that both components.css coming from https://assets.finn.no/pkg/@warp-ds/css/v1/components.css and this file are always in sync, because they would both point to the same version of the css package on Eik. Wdyt?

BalbinaK (commented):
Really great job on this one! I just have a few comments and an open question :) 

> consumers will need to add an async wrapper :(

Hm. Just an open question. Would it help if we exported `getClassesToPurge` function from this package and told users to pass its result to `externalClasses` instead of adding the `omitComponentClasses` flag? Might be they'd still need to add an async wrapper to the defineConfig() though.

AnnaRybkina:
> Really great job on this one! I just have a few comments and an open question :)
> 
> > consumers will need to add an async wrapper :(
> 
> Hm. Just an open question. Would it help if we exported `getClassesToPurge` function from this package and told users to pass its result to `externalClasses` instead of adding the `omitComponentClasses` flag? Might be they'd still need to add an async wrapper to the defineConfig() though.

Yea, if I understand the suggestion correctly the consumers would need to add a wrapper as it is for the fetching that is async. In that way to only add a wrapper seems more convenient and easy for the consumers to use. But maybe I did not understand correctly

src/classesToPurge.js
AnnaRybkina: Yes, it would be good. We need to merge the css PR to publish the classes and then use them here. Should not be a problem

test/_helpers.js
AnnaRybkina: True, I haven't thought about this one. Good catch. We can safely remove this one

dev.js
AnnaRybkina: Yes, I was testing this back and forward and left it like this by mistake. I pushed the fix for this now. I'll test locally when we have a meeting. It was confusing to leave it like that, sorry

BalbinaK:
> Yea, if I understand the suggestion correctly the consumers would need to add a wrapper as it is for the fetching that is async. In that way to only add a wrapper seems more convenient and easy for the consumers to use. 

Now that I thought about it again, you're right that it would be a bit easier for users to add an async wrapper to their config instead of also handling the fetching of classes. I guess the reason why I thought about it was that it would maybe be more apparent for them why the async wrapper is at all needed. But then again, we'd at least always control the eik url for classesToPurge.json, if for any reason that url was to be replaced with something else.

src/classesToPurge.js
```diff
@@ -0,0 +1,17 @@
+let classes;
+async function getClassesToPurge() {
+  if (classes) return classes;
+  else {
+    classes = await fetch(
+      'https://assets.finn.no/pkg/@warp-ds/css/2.0.0-v2.1/classesToPurge.json',
```
AnnaRybkina, line 6: To be changed to v2 before merge
```suggestion
      'https://assets.finn.no/pkg/@warp-ds/css/v2/classesToPurge.json',
```

dev.js
```diff
@@ -22,8 +21,11 @@ const {
     },
   },
 });
-
-const uno = createGenerator({ presets: [presetWarp( { ...options, development: true } )] });
+const asyncPreset = await presetWarp({
+  ...options,
+  development: true,
+});
+const uno = createGenerator({ presets: [asyncPreset] });
```
BalbinaK, line 28: I'm still wondering if this particular script should still include all classes, also those that are part of warp.css (current components.css). We'd probably need to pass those `classesToPurge` to safelist here then. 
It's just that if we don't generate css for classes that are always externalized in this script or in tests, we won't see if our rules still work correctly, and those are used in @warp-ds/css to generate that warp.css stylesheet. Maybe there's a different way to solve it so it's more of a question what you think about it.

AnnaRybkina: If `development: true` is set we will generate css for any class you want to test. Or do you mean something else?

BalbinaK: Ah, of course. Sorry!

BalbinaK (approved):

AnnaRybkina:
Closing this, read more here : https://sch-chat.slack.com/archives/C050N3NQ204/p1698311845906879

PR #170, opened 2023-11-08
feat: add variantPartClasses to _variants

felicia-haggqvist:
Add `variantPartClasses` to _variants

**Note:** This is part of implementing changes to the expandable-component in the elements-repo in order to be able to style the svg from a warp-icon that is inside a shadow-DOM using the [CSS pseudo-element part](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) that is also supported by unoCSS.

### DIFF
--- a/src/_variants/index.js
+++ b/src/_variants/index.js
@@ -5,6 +5,7 @@ import {
   variantTaggedPseudoClasses,
   variantPseudoClassFunctions,
   variantPseudoClassesAndElements,
+  variantPartClasses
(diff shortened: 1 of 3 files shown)


BalbinaK (approved):
Would be great if you could add a test case to `/test/variants.js` just to make sure the generated CSS is as expected when we e.g. update @unocss/preset-mini dependency in the future. That also can help us understand what the given change means in practice :)

PR #173, opened 2023-11-22
feat(rules): add support for outline color classes

BalbinaK:
Fixes [WARP-398](https://nmp-jira.atlassian.net/browse/WARP-398):

We should allow users to style the color of the outline in the same way as they can style border - using semantic color classes. We can for the time being reuse border color tokens, but since the focused outline color is currently assigned to a generic `--w-s-color-focused` token, we thought it’s better to have it as  `--w-s-color-border-focused` to make the border and outline color rules aligned. 

We're keeping the generic `--w-s-color-focused` token for backward compatibility, but we can indicate that it may be removed by setting the `--w-s-color-border-focused` on `focusable` with the previous CSS variable as fallback. I'm open to solving it in a different way. We haven't really spoken about the future of `--w-s-color-focused` 😄

### DIFF
--- a/src/_rules/semantic.js
+++ b/src/_rules/semantic.js
@@ -15,5 +15,6 @@ export const semanticRules = [
   [/^s-border-([lrtbxy])-(.+)$/, ([, direction, cssvar]) => directionMap[direction]?.map(
     (dir) => [`border${dir}-color`, h.semanticToken(`border-${cssvar}`)],
   )],
+  [/^s-outline$/, () => ({ 'outline-color': h.semanticToken('border') })],
   [/^s-outline-(.+)$/, ([, cssvar]) => ({ 'outline-color': h.semanticToken(`border-${cssvar}`) })],
 ];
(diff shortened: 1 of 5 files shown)


felicia-haggqvist (approved):

src/_rules/semantic.js
```diff
@@ -15,4 +15,5 @@ export const semanticRules = [
   [/^s-border-([lrtbxy])-(.+)$/, ([, direction, cssvar]) => directionMap[direction]?.map(
     (dir) => [`border${dir}-color`, h.semanticToken(`border-${cssvar}`)],
   )],
+  [/^s-outline-(.+)$/, ([, cssvar]) => ({ 'outline-color': h.semanticToken(`border-${cssvar}`) })],
```
magnuh, line 18: Do we also need this:
`[/^s-outline$/, () => ({ 'outline-color': h.semanticToken('border') })]`
to be able to set the color to `s-border`? 🤔

BalbinaK: Ah, you might be right! I'll just double-check with @Skadefryd.

Skadefryd: Yes I guess :) Im not 100% sure I understand the comment but I think we should follow the same patterns as much as possible (when it makes sense)   Outline and border should be mostly the same from the consumers standpoint.

Skadefryd (approved):

src/_rules/semantic.js
BalbinaK: Added support for `s-outline` in [e5c9331](https://github.com/warp-ds/drive/pull/173/commits/e5c9331e3347f5300d1dbcfa97a7dec7907c9cbc)

PR #176, opened 2023-12-07
feat: Add missing support for arbitrary and semantic colors

magnuh:
- Add support for css variables in arbitrary values for border and divide
- Add support for css variables in arbitrary values for text
- Add support for color and css variables in arbitrary values for outline
- Add support for semantic classes setting divide color

Co-authored by: @Skadefryd

### DIFF
--- a/src/_rules/color.js
+++ b/src/_rules/color.js
@@ -12,6 +12,7 @@ export const textColors = [
 //  ['text-inherit', { 'color': 'inherit' }], // This class currently sets "text-align: inherit;" in align.js
   ['text-transparent', { 'color': 'transparent' }],
   ['text-current', { 'color': 'currentColor' }],
+  [/^text-\[(.+)]/, ([, p]) => ({ 'color': p.startsWith('var') ? p : `var(${p})` })],
 ];
 
 export const bgColors = [
--- a/src/_rules/outline.js
+++ b/src/_rules/outline.js
@@ -5,6 +5,13 @@ const outlineNone = {
   'outline-offset': '2px',
 };
 
+export const outlineColors = [
+  ['outline-inherit', { 'outline-color': 'inherit' }],
+  ['outline-transparent', { 'outline-color': 'transparent' }],
+  ['outline-current', { 'outline-color': 'currentColor' }],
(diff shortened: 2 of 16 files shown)


felicia-haggqvist (approved):

felicia-haggqvist (approved):

src/_rules/color.js
```diff
@@ -12,6 +12,7 @@ export const textColors = [
 //  ['text-inherit', { 'color': 'inherit' }], // This class currently sets "text-align: inherit;" in align.js
   ['text-transparent', { 'color': 'transparent' }],
   ['text-current', { 'color': 'currentColor' }],
+  [/^text-\[(.+)]/, ([, p]) => ({ 'color': p.startsWith('var') ? p : `var(${p})` })],
```
BalbinaK, line 15: Since we use `^` to match the beginning, shouldn't there be `$` after `]` to mark the end of the string?

src/_rules/outline.js
```diff
@@ -5,6 +5,13 @@ const outlineNone = {
   'outline-offset': '2px',
 };
 
+export const outlineColors = [
+  ['outline-inherit', { 'outline-color': 'inherit' }],
+  ['outline-transparent', { 'outline-color': 'transparent' }],
+  ['outline-current', { 'outline-color': 'currentColor' }],
+  [/^outline-\[(.+)]/, ([, p]) => ({ 'outline-color': p.startsWith('var') ? p : `var(${p})` })],
```
BalbinaK, line 12: same here about `$` at the end of the expression

src/_rules/color.js
BalbinaK, line 15: Have you also discussed accepting arbitrary values for text size? We have support for arbitrary values for line-height in [/_rules/typography.js](https://github.com/warp-ds/drive/blob/next/src/_rules/typography.js#L12):
```
[/^leading-\[(.+)(rem|px)?\]/, ([, value, unit], context) => ({ 'line-height': resolveArbitraryValues(value, unit, context) })]
```

If it was considered to target classes with `/^text-\[(\d+)(rem|px)?\]$` then the rule like `/^text-\[(.+)]/` might take over.

PR #180, opened 2023-12-22
feat: support arbitrary values for flex-basis, flex-grow, flex-shrink, grid-auto-rows, grid-auto-columns

felicia-haggqvist:
Fixes JIRA ticket: [WARP-412](https://nmp-jira.atlassian.net/browse/WARP-412)

- Added support for arbitrary values for flex-grow and flex-shrink, following the unit-less convention, for example: `shrink-[5]`, `grow-[4]`
- Added support for arbitrary values for flex-basis, follows a unit-less + unit convention, for example: `basis-[32]` ,`basis-[15px]`
- Added support for arbitrary value for grid-auto-rows, grid-auto-columns, following both, unit + unit-less convention, underscore convention + letter + parenthesis, for example: `auto-rows-[minmax(0,2fr)]`, `auto-cols[20px_30px_0%]`
- Updated `content-none` to set `{content: none}` instead of `’””’`.
- Added tests for arbitrary values for flex-basis, flex-grow, flex-shrink, grid-auto-rows, grid-auto-columns, background-position (currently commented out), content-none, content-empty, and content-[””]

**Note 1:**
Not sure if the refactoring of `content-none` would mean a breaking change? 

**Note 2:**
Wait with adding support to arbitrary values for background-position until changes from JIRA ticket [WARP-415](https://nmp-jira.atlassian.net/browse/WARP-415) has been merged. Need to find a way to not break support for arbitrary values for background-color since both background-position and background-color start with `bg-`.

### DIFF
--- a/src/_rules/background.js
+++ b/src/_rules/background.js
@@ -1,4 +1,4 @@
-import { positionMap, globalKeywords, makeGlobalStaticRules } from '#utils';
+import { positionMap, globalKeywords, makeGlobalStaticRules, handleComplexArbitraryValues } from '#utils';
 
 export const backgrounds = [
   // size
@@ -37,13 +37,16 @@ export const backgrounds = [
   ['bg-origin-content', { 'background-origin': 'content-box' }],
   ...makeGlobalStaticRules('bg-origin', 'background-origin'),
 
-  //arbitrary
-  [/^bg-\[(.+)]/, ([, p]) => {
+  // matching arbitrary values
+  [/^bg-\[(.+)]/, ([, p, unit, context]) => {
     if (p.startsWith('url')) {
       // Process url(var(--something)) and extract the var itself -> --something
       const urlAsVar = p.match(/url\(var\(([^)]*)/)?.[1];
       return { 'background-image': urlAsVar ? `var(${urlAsVar})` : p };
+    } else if (p.startsWith('var')) {
+      return { 'background-color': p.startsWith('var') ? p : `var(${p})` };
+    } else {
+      return {'background-position': handleComplexArbitraryValues(p, unit, context)}; 
     }
-    return { 'background-color': p.startsWith('var') ? p : `var(${p})` };
   }],
 ];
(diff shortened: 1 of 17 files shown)


src/_rules/background.js
```diff
@@ -37,13 +37,16 @@ export const backgrounds = [
...
-  //arbitrary
-  [/^bg-\[(.+)]/, ([, p]) => {
+  // matching arbitrary values
+  [/^bg-\[(.+)]/, ([, p, unit, context]) => {
     if (p.startsWith('url')) {
       // Process url(var(--something)) and extract the var itself -> --something
       const urlAsVar = p.match(/url\(var\(([^)]*)/)?.[1];
       return { 'background-image': urlAsVar ? `var(${urlAsVar})` : p };
+    } else if (p.startsWith('var')) {
+      return { 'background-color': p.startsWith('var') ? p : `var(${p})` };
```
magnuh, line 47: This won't work as it will only enter this if clause when p starts with "var". 😄
I have this covered in my branch (with color/variable fixes) though... 😄 So you could just remove this part for now. 🙂

src/_rules/background.js
```diff
@@ -37,13 +37,16 @@ export const backgrounds = [
   ['bg-origin-content', { 'background-origin': 'content-box' }],
   ...makeGlobalStaticRules('bg-origin', 'background-origin'),
 
-  //arbitrary
-  [/^bg-\[(.+)]/, ([, p]) => {
+  // matching arbitrary values
+  [/^bg-\[(.+)]/, ([, p, unit, context]) => {
```
magnuh, line 41: This does not look right? 🤔 It has only one matching group, so unit won't work... and context is supposed to go as the second parameter for the arrow function and not in the first array of matching groups. 🙂

magnuh:
> **Note:** Not sure if the refactoring of `content-none` would mean a breaking change?

Even though it might break for anyone that has used it, I would still consider it a bug fix, since it was not working as expected before. 😄 (Tailwind also sets it to `content: none`)

src/_rules/background.js
felicia-haggqvist: Oh you’re right! 🙈

magnuh (approved):

PR #181, opened 2023-12-29
feat: support arbitrary values for order, columns and transition-timing

felicia-haggqvist:
Fixes JIRA ticket: [WARP-413](https://nmp-jira.atlassian.net/browse/WARP-413)

- Adjust rule for order to only accept 1-12, and add support for arbitrary values for order following the unit-less convention, for example: `order-[14]`

- Add support for columns (accept 1-12), as well as, support arbitrary values for columns, following the unit-less convention, for example: `columns-[13]`

- Add support for arbitrary values for transition-timing function, following a unit-less convention together with `cubic-bezier`, for example: `ease-[cubic-bezier(0.95,0.05,0.795,0.035)]`

- Add tests for order, columns and transition-timing

### DIFF
--- a/src/_rules/layout.js
+++ b/src/_rules/layout.js
@@ -21,15 +21,14 @@ export const overflows = [
 ];
 
 export const columns = [
-  [
-    /^columns-([1-9]|1[0-2])$/,
+  ['columns-auto', { 'columns': 'auto' }],
+  
+  [/^columns-([1-9]|1[0-2])$/,
     ([, d]) => ({ 'columns': h.number(d) }),
     { autocomplete: 'columns-<num>' },
   ],
-  ['columns-auto', { 'columns': 'auto' }],
   // matching arbitrary values
-  [
-    /^columns-\[(\d+)?]$/,
+  [/^columns-\[(\d+)?]$/,
     ([, d]) => ({ 'columns': h.number(d) }),
     { autocomplete: 'columns-<num>' },
   ],
--- a/src/_rules/layout.js
+++ b/src/_rules/layout.js
@@ -28,7 +28,7 @@ export const columns = [
     { autocomplete: 'columns-<num>' },
   ],
   // matching arbitrary values
-  [/^columns-\[(\d+)?]$/,
+  [/^columns-\[(\d+)]$/,
     ([, d]) => ({ 'columns': h.number(d) }),
     { autocomplete: 'columns-<num>' },
   ],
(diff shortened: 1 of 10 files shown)


src/_rules/layout.js
```diff
@@ -17,3 +19,18 @@ export const overflows = [
...
+export const columns = [
+  [
+    /^columns-(\d+)$/,
+    ([, d]) => ({ 'columns': h.number(d) }),
+    { autocomplete: 'columns-<num>' },
+  ],
+  ['columns-auto', { 'columns': 'auto' }],
+  // matching arbitrary values
+  [
+    /^columns-\[(.\d*)?]$/,
```
magnuh, line 32: Should this be something like 
```suggestion
    /^columns-\[(\d\d*)?]$/,
```
to not accept things like `columns-[z33]`? 🤔

src/_rules/layout.js
```diff
@@ -17,3 +19,18 @@ export const overflows = [
...
+export const columns = [
+  [
+    /^columns-([1-9]|1[0-2])$/,
+    ([, d]) => ({ 'columns': h.number(d) }),
+    { autocomplete: 'columns-<num>' },
+  ],
+  ['columns-auto', { 'columns': 'auto' }],
+  // matching arbitrary values
+  [
+    /^columns-\[(\d*)?]$/,
```
magnuh, line 32: Should we allow empty brackets here? Or should it be changed to:
```suggestion
    /^columns-\[(\d+)]$/,
```

felicia-haggqvist: no you are right! We probably don't want to allow empty brackets. 😄 I'll change it!

src/_rules/position.js
```diff
@@ -7,13 +7,19 @@ export const positions = [
...
+    /^order-([1-9]|1[0-2])$/,
     ([, d]) => ({ 'order': h.number(d) }),
     { autocomplete: 'order-<num>' },
   ],
   ['order-first', { order: '-9999' }],
   ['order-last', { order: '9999' }],
   ['order-none', { order: '0' }],
+  // matching arbitrary values
+  [
+    /^order-\[(\d*)?]$/,
```
magnuh, line 19: Same here I guess? 🙂 
```suggestion
    /^order-\[(\d+)]$/,
```
...or if we want to allow empty brackets we could just remove the question mark in the end `/^order-\[(\d*)]$/` 🙂

felicia-haggqvist: I don't see why a user should just add an empty bracket, so maybe it's best if I change it so that we don't allow it :)

magnuh (approved):

PR #186, opened 2024-01-10
fix: Improve arbitrary values handling for border and divide

magnuh:
Fixes handling of all types of supported arbitrary values for `border` and `divide`

- Values starting with `--` will be setting border-color (transcribed as `var(--...)`)
- Values added as `var(...)` will set `border-color`
- Values added as numbers followed by optional unit will set `border-width`
- Any other arbitrary values will set `border-color` (e.g. "#000000", "black", "rgb(0,0,0)", "asdfasdf"...)

This should also work with any valid direction ([lrtbxy] for border, [xy] for divide).


felicia-haggqvist (approved):
LGTM! 🚀

PR #194, opened 2024-01-23
feat: Add support for arbitrary box-shadow

magnuh:


src/_rules/slider.js
```diff
@@ -1,5 +1,5 @@
-// Hardcoded styling for slider border shadow, will be modified and handled with correct color variables at a later stage
-// TODO: Make the colors for the slider shadow themeable
+// Deprecated - DO NOT USE! - Replace with arbitrary box-shadow (box-[---token]) to handle theming
```
BalbinaK, line 1:
```suggestion
// Deprecated - DO NOT USE! - Replace with arbitrary box-shadow (shadow-[--token]) to handle theming
```

magnuh: Oops 😄 🙏

PR #197, opened 2024-02-01
feat: Add support for alpha channel modifier to semantic color classes

magnuh:
Adds support for an optional alpha channel modifier to semantic color classes (e.g. `s-bg-primary/50`, `s-border/25` ...).

Requires changes in:
https://github.com/warp-ds/tokenizer/pull/2
https://github.com/warp-ds/css/pull/161


imprashast (approved):
This is so nice 👍

PR #198, opened 2024-02-05
feat: Add support for alpha channel modifier to arbitrary color classes

magnuh:
Adds support for an optional alpha channel modifier to arbitrary color classes (e.g. `bg-[--w-s-border-primary/50]`, `border-[var(--w-black)/25]` ...).


felicia-haggqvist (approved):
LGTM! :)

PR #206, opened 2024-02-27
chore: add caching to CI test runs

pearofducks:


BalbinaK:
Just curious, what was the reason this caching was needed?

Also, this PR needs rebasing after a release commit was pushed.

pearofducks:
Caching is considered good practice in general:
- it hits NPM less often, so uses less bandwidth
- it usually speeds up CI runs to use locally cached assets

Doesn't really need rebasing if squash is used on the PR.

PR #215, opened 2024-04-15
fix: New svg icon and styling fixes for list-checked

magnuh:
image

Uses this: https://github.com/warp-ds/css/pull/194


Skadefryd (approved):
Nice and simple fix, gj!

PR #216, opened 2024-04-15
feat: Added rules for break-inside

magnuh:
Adds missing rules for: https://warp-ds.github.io/css-docs/break-inside#setting-the-break-inside-behavior

Reported here: https://sch-chat.slack.com/archives/C04P0GYTHPV/p1713167639919199


felicia-haggqvist (approved):

Skadefryd:
Fixed docs for that, might have to be version bumped for the qr table on top to get the new supported stuff(?)  - >     https://github.com/warp-ds/css-docs/pull/107

PR #219, opened 2024-04-23
feat: support css vars as arbitrary values for margins and paddings

BalbinaK:
### Changes included
CSS variables are now valid arbitrary values for padding and margin. This will allow setting a class on Warp TextFields with long prefixes like `pl-[var(--w-prefix-width,_40px)]`  instead of using [inline css](https://github.com/warp-ds/elements/blob/next/packages/textfield/index.js#L43-L45) or [style blocks](https://github.com/warp-ds/react/blob/next/packages/textfield/src/component.tsx#L48-L54).

Additionally, this PR contains the following changes:
- [fix: remove bracket handler from directionSize util](https://github.com/warp-ds/drive/commit/b321b389630decb6d1a902576678f241de3fe42a) - `bracket()` handler for matching arbitrary values was removed from `directionSize` util as we are handling all arbitrary values as separate rules now. This helped specify more strict arbitrary rules, as invalid values were no longer handled in the more "forgiving" rules which relied on the `directionSize` util.
- [fix: padding again supports global/auto values](https://github.com/warp-ds/drive/commit/5e8afb941ff408c02471b7a116f96d5cac57d0af) - padding rules were adjusted to again support auto and global values (the `directionSize` util handles those)

Co-autherd by: @magnuh

### DIFF
--- a/src/_rules/spacing.js
+++ b/src/_rules/spacing.js
@@ -6,28 +6,28 @@ export const padding = [
   [/^p()-(.\d*)$/, directionSize('padding'), { autocomplete: '(m|p)-<num>' }],
   [/^p([xy])-(.\d*)$/, directionSize('padding')],
   [/^p([rltb])-(.\d*)$/, directionSize('padding'), { autocomplete: '(m|p)<directions>-<num>' }],
-  // matching arbitrary values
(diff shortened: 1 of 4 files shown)


felicia-haggqvist (approved):
LGTM! 🎉

PR #222, opened 2024-05-16
feat: Added leading-none rule to set line-height to 1

magnuh:
Added missing rule for setting line-height to 1:
`.leading-none:{line-height:1}`

### DIFF
--- a/src/_rules/typography.js
+++ b/src/_rules/typography.js
@@ -3,6 +3,7 @@ import { textMap, lineHeightMap, resolveArbitraryValues } from '#utils';
 export const typography = [
   [/^text-(12|14|16|20|22|28|34|48)$/, ([, d]) => ({ 'font-size': `var(--w-font-size-${textMap[d]})`, 'line-height': `var(--w-line-height-${textMap[d]})` })],
   [/^text-(xs|s|m|ml|l|xl|xxl|xxxl)$/, ([, size]) => ({ 'font-size': `var(--w-font-size-${size})`, 'line-height': `var(--w-line-height-${size})` })],
+  ['leading-none', { 'line-height': 1 }],
   [/^leading-(16|18|22|26|28|34|41|56)$/, ([, d]) => ({ 'line-height': `var(--w-line-height-${lineHeightMap[d]})` })],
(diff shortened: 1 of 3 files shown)


BalbinaK (commented):
Some context for this change would be nice, like where did you see we needed to add this particular utility class. Maybe there was a discussion I missed - then sorry :D I see in TailwindCSS they also use `leading-loose` which applies `line-height: 2` - do we want that, too?

magnuh:
> Some context for this change would be nice, like where did you see we needed to add this particular utility class. Maybe there was a discussion I missed - then sorry :D I see in TailwindCSS they also use `leading-loose` which applies `line-height: 2` - do we want that, too?

We have t-shirt sizes for that (leading-xs, leading-s...).
leading-none is needed for setting the line-height to the same as the font height... Which you often have to do in order to align things. The absence of this class is the reason for style="line-height:1" being sprinkled all over the css/tech-docs. 😅

BalbinaK:
> > Some context for this change would be nice, like where did you see we needed to add this particular utility class. Maybe there was a discussion I missed - then sorry :D I see in TailwindCSS they also use `leading-loose` which applies `line-height: 2` - do we want that, too?
> 
> We have t-shirt sizes for that (leading-xs, leading-s...). leading-none is needed for setting the line-height to the same as the font height... Which you often have to do in order to align things. The absence of this class is the reason for style="line-height:1" being sprinkled all over the css/tech-docs. 😅

I’m not sure about the `leading-none` still. No one has ever requested it and is sounds like a typography change that could go against what our design system allows 🤔 I’d rather request @Skadefryd's review and/or double check with designers if we want to allow a line height that is the same as the font size.

magnuh:
> I’m not sure about the `leading-none` still. No one has ever requested it and it sounds like a typography change that could go against what our design system allows 🤔 I’d rather request @Skadefryd's review and/or double check with designers if we want to allow a line height that is the same as the font size.

Sure. Meanwhile everyone, including ourselves, will just stick to `style="line-height:1"` to reset the line-height. 👍

felicia-haggqvist (approved):

PR #223, opened 2024-05-21
fix: Tweak positioning of list-checked icon

magnuh:
Tweak positioning of list-checked icon to work as good as possible with different font-sizes/line-heights.

Reported here: https://nmp-jira.atlassian.net/browse/WARP-512?focusedCommentId=224873

### DIFF
--- a/src/_rules/list.js
+++ b/src/_rules/list.js
@@ -28,7 +28,7 @@ export const listStyle = [
 const listCheckedItemStyles = entriesToCss(
   Object.entries({
     position: 'relative',
-    'padding-left': '1.2em',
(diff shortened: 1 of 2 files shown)


BalbinaK (commented):
Nice! Have you checked if it looks good with Tori fonts, too? Also, I see tests are failing now so probably an update to a snapshot is required :)

magnuh:
> Nice! Have you checked if it looks good with Tori fonts, too?

It looks ok for Tori for now... will be better if/when the baseline of the Tori font is adjusted.

BalbinaK (approved):

PR #230, opened 2024-12-02
fix: pin to uno 0.64.1

wkillerud:
Hey gang.

[Uno 0.65.0](https://github.com/unocss/unocss/releases/tag/v0.65.0) has a breaking change which affects us. With the dependencies as they are now we end up with the latest `0.x` by default. It's not semver stable, so this PR pins the unocss version to the last known working version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,9 @@
   "author": "@djh",
   "license": "Apache-2.0",
   "dependencies": {
-    "@unocss/core": "0.x",
-    "@unocss/rule-utils": "0.x",
-    "@unocss/preset-mini": "0.x"
+    "@unocss/core": "0.64.1",
+    "@unocss/rule-utils": "0.64.1",
+    "@unocss/preset-mini": "0.64.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


imprashast (approved):

imprashast:
Remember to revert this once web-platform team has upgraded their stuff to UnoCSS@0.65.0

package.json
```diff
@@ -66,5 +66,6 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
-  }
+  },
+  "packageManager": "pnpm@9.12.1+sha512.e5a7e52a4183a02d5931057f7a0dbff9d5e9ce3161e33fa68ae392125b79282a8a8a470a51dfc8a0ed86221442eb2fb57019b0990ed24fab519bf0e1bc5ccfc4"
```
BalbinaK, line 70: @imprashast it may not be relevant anymore but I remember us having issues with CI or something when we specified a version in `packageManager` - we removed that e.g. from @warp-ds/css in https://github.com/warp-ds/css/pull/208

imprashast: Let's see 🤞

PR #242, opened 2025-03-17
feat: allow targeting using data attribute like `data-[is-active]:s-bg-primary`

Gyllsdorff:
```tsx
const Link = ({title, isActive, path}: {isActive?: Boolean; title: string; path: string}) => (
	<li className="inline-block hover:s-bg-hover">
		<a
			data-is-active={isActive}
			className="inline-block text-body font-bold p-16 s-text data-[is-active]:s-bg-negative"
			href={path}
		>
			{title}
		</a>
	</li>
)
```

### DIFF
--- a/test/__snapshots__/variants.js.snap
+++ b/test/__snapshots__/variants.js.snap
@@ -39,6 +39,8 @@ exports[`variants > negative 1`] = `
 exports[`variants > pseudo 1`] = `
 "/* layer: default */
 .focus\\:border:focus{border-width:1px;}
+.aria-\\[invalid\\=spelling\\]\\:s-bg-primary[aria-invalid=spelling],
+.data-\\[is-active\\]\\:s-bg-primary[data-is-active],
 .part-\\[part-name\\]\\:s-bg-primary::part(part-name){background-color:var(--w-s-color-background-primary);}
 .group:hover .group-hover\\:p-16{padding:1.6rem;}
 .before\\:p-32::before{padding:3.2rem;}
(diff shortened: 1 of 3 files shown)


test/__snapshots__/variants.js.snap
```diff
@@ -39,6 +39,8 @@ exports[`variants > negative 1`] = `
 exports[`variants > pseudo 1`] = `
 "/* layer: default */
 .focus\\:border:focus{border-width:1px;}
+.aria-\\[invalid\\=spelling\\]\\:s-bg-primary[aria-invalid=spelling],
+.data-\\[is-active\\]\\:s-bg-primary[data-is-active],
```
Gyllsdorff, line 43: Hmm, are these test really testing the right thing? 🤔

Gyllsdorff: Ah, it is because I used the same color as the other selector, that why there is a `,` at the end. 😅

test/variants.js
```diff
@@ -30,6 +30,8 @@ describe('variants', () => {
       'group/llama',
       'group-hover/llama:m-32',
       'part-[part-name]:s-bg-primary',
+      'data-[is-active]:s-bg-primary',
+      'aria-[invalid=spelling]:s-bg-primary',
```
imprashast, line 34: this test doesn't produce any valid css. Suggest to add 2 separate tests like:

 ```
 test('data variant', async ({ uno }) => {
    const classes = ['data-[is-active]:s-bg-primary'];
    const { css } = await uno.generate(classes);
    expect(css).toMatchSnapshot();
  });

  test('aria variant', async ({ uno }) => {
    const classes = ['aria-[invalid=spelling]:s-bg-primary'];
    const { css } = await uno.generate(classes);
    expect(css).toMatchSnapshot();
  });
```

Gyllsdorff: 👍 Done.

imprashast (approved):

